PR #1398, opened 2014-11-25
Removed extra version check

t0xicCode:
Shinken requires version 2.6, so it is unnecessary to check if the python version is higher or equal to 2.6, as it will always be true.

This PR removes that extra version check.


naparuba:
good ones, thanks :)

PR #1408, opened 2014-12-04
Let's change the included files in the bin/ hierarchy. We want to includ...

ofosos:
...e all files

under both init.d and rc.d for the source distribution. While the rc.d files are
for BSD, the init.d files are for Linux distributions. When installing a source
distribution this will install the data_files (cf. setup.py) needed for the specific
target the we are installing on.

This should fix pip on FreeBSD.


naparuba:
Seems to be done. I close. Feel free to reopen if need ^^

PR #1410, opened 2014-12-08
return_code value is always 0

jccomputing:
When broker sends a host_check_result or service_check_result brok, the return_code value is always 0 (default value).

The return_code is interesting for archival and proof purpose, so I set it the the exit_status value.

I tested and it works fine. I hope that's the correct way to do this.

### DIFF
--- a/shinken/objects/schedulingitem.py
+++ b/shinken/objects/schedulingitem.py
@@ -848,6 +848,9 @@ class SchedulingItem(Item):
 
         self.set_state_from_exit_status(c.exit_status)
 
+        # Set return_code to exit_status to fill the value in broks
+        self.return_code = c.exit_status
+
         # we change the state, do whatever we are or not in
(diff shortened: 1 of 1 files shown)


naparuba:
I don't understand, what i the role of self.return_code ? I do'nt see any usage of it

jccomputing:
I understand it sounds a little weird. I myself wondered when I was asked to do it.

It's for archival purpose, when someone challenges you, tells that Shinken didn't poll correctly and you have to prove the opposite.

It's the only data missing in broks. We aleady have the state, the state_type, the output, the perfdatas, etc.

gst:
I agree (s|h)aving the exit-status/return-code of plugins/probes can be useful..
after you have to see where to use it eventually. (logs ?)

naparuba:
Thanks :)

PR #1415, opened 2014-12-10
Ensure that livestatus client close its connections

rledisez:
The LSConnectionPool does not close the connections after the end of the
requests. On long running process doing regular livestatus request, it
quickly saturate the maximum allowed file descriptor. The commit ensures
that connections are closed after the results have been read.


naparuba:
Thanks :)

I did not know someone did use this lib in fact :)

PR #1423, opened 2014-12-15
Use dedicated setup_test.sh for easier test setup

gst:
Created dedicated setup_test.sh script to be used by shinken modules for their tests.
Also make use of it for shinken itself.


naparuba:
Thanks ^^

gst:
good :p 
Now we can use it by shinken modules for setting up their test env ! (when they need shinken itself, which they all do as far as know for now)
instead of duplicating things like that :p

PR #1428, opened 2014-12-23
[WIP] Rework setup.py (virtualenv-compatible)

titilambert:
Hello !
We have begun to rework the setup.py to be virtualenv compatible
Of course, this could/should be merge after the 2.2 release...

We are currently fixing tests, but most of it works...


bin/preparedev
```diff
@@ -1,52 +0,0 @@
...
-
-# If the share modules is missing (dir or link) create a link with our share dir
-if [ ! -L /var/lib/shinken/share ] && [ ! -d /var/lib/shinken/share ]; then
-  # Control will enter here if $DIRECTORY doesn't exist.
-  sudo ln -s  $HERE/../share /var/lib/shinken/share
-fi
-
-sudo chown $ME:$ME /var/run/shinken
-
-
```
ReAzem: Shouldn't we just fix this script? Cleaning logs seems useful.
If we decide not to keep it, then we should remove it from the ```bin/launch_*.sh``` scripts

titilambert:
> Btw, what kind of magic are you using to make it work on bsd? You know path are /usr/local/***. Setup tools are dealing with it?

@Seb-Solon I think it's working on freeBSD, cause python setup.py install default prefix path is /usr/local/***  :)

titilambert:
Also, If you want to install with root user, you have to use : sudo python setup.py install --prefix=/

install_hooks.py
```diff
@@ -0,0 +1,60 @@
+import os 
+import re
+import fileinput
+
+
+def main(config):
+    """ function called before any actions in python setup.py install """
+    print "ASAAAAAAAAA"
```
ppepos: Maybe we can remove this XD

titilambert:
This PR was rebased here: #1467

PR #1435, opened 2015-01-06
Shinken notification

claneys:
Here is the pull request about new email notify plugin.


etc/shinken.cfg
```diff
@@ -79,7 +79,7 @@ disable_old_nagios_parameters_whining=0
 #use_timezone=Europe/Paris
 
 # Disabling env macros is good for performances. If you really need it, enable it.
-enable_environment_macros=0
+enable_environment_macros=1
```
naparuba: oups oups oups: this is a very bad thing to enable globaly, really. This made communication and CPU explode on schedulers.

titilambert:
@Seb-Solon +1
This is a plugin which could be run with nagios/icinga/... 
@naparuba What do you think about including plugins in Shinken core ?

claneys:
@Seb-Solon : I added a parameter to include or not a webui link. But I don't see how to get that link withtout shinken.config.Config.

Anyway, this is plugin no doubt about it but has we already have some plugins shipped with shinken, what's the problem ? This one is like notify_by_xmpp.py or external_mapping.py, I don't see the difference.

Seb-Solon:
The config is just an python object , you can get the webui link by reading the file directly or sking a parameter. Here you are using shinken.config.Config as a lib, which is a bit overkill for that. 

Shinken core should not provide plugin actually, we should consider removing the other ;)

claneys:
@Seb-Solon did you have the time to review the code ? I can help if needed. Thanks

claneys:
Didn't see your comment. Ok for the use of shinken.config object, I will change that, thanks.

claneys:
Where these plugins has to go ?

I did not find any repos about basics shinken plugins in shinken-monitoring github account.

What about put them in a pack that can be installed by shinken cli command ?

Seb-Solon:
The common practice on the shinken-monitoring is to ship plugins with pack. So yeah you could put that with a pack. I close this as we can agree it's not shinken-core related.

PR #1459, opened 2015-01-15
tests: clean Unittest2CompatMixIn: not anymore needed since we use unitt...

gst:
...est2 with py26


gst:
can be merged directly as it had no more effect since we moved to unittest2 (for python2.6)

PR #1463, opened 2015-01-19
Fix template self use

gst:
+ Prevent infinite recursion (and so crash) if a template use itself -> configuration error.
+ Report (at leasta ) configuration warning if an item use an unknown template.

### DIFF
--- a/shinken/objects/item.py
+++ b/shinken/objects/item.py
@@ -709,6 +709,14 @@ Like temporary attributes such as "imported_from", etc.. """
                 dmp[prop] = getattr(self, prop)
         return dmp
 
+    def _get_name(self):
+        if hasattr(self, 'get_name'):
+            return self.get_name()
+        name = getattr(self, 'name', None)
+        host_name = getattr(self, 'host_name', None)
+        return '%s(host_name=%s)' % (name or 'no-name', host_name or '')
+
+
 
 class Items(object):
     def __init__(self, items, index_items=True):
@@ -992,6 +1000,7 @@ class Items(object):
             all_tags.extend(self.get_all_tags(t))
         return list(set(all_tags))
 
+
     def linkify_item_templates(self, item):
         tpls = []
         tpl_names = item.get_templates()
@@ -1002,13 +1011,13 @@ class Items(object):
                 # TODO: Check if this should not be better to report as an error ?
                 self.configuration_warnings.append(
                     "%s %r use/inherit from an unknown template (%r) ! Imported from: %s" % (
-                        type(item).__name__, item.name, name, item.imported_from                                                                                              )
-                )
+                        type(item).__name__, item._get_name(), name, item.imported_from
+                ))
             else:
                 if t is item:
                     self.configuration_errors.append(
                         '%s %r use/inherits from itself ! Imported from: %s' % (
-                            type(item).__name__, item.name, item.imported_from
+                            type(item).__name__, item._get_name(), item.imported_from
                     ))
                 else:
                     tpls.append(t)
--- a/shinken/objects/item.py
+++ b/shinken/objects/item.py
@@ -998,7 +998,13 @@ class Items(object):
 
         for name in tpl_names:
             t = self.find_tpl_by_name(name)
-            if t is not None:
+            if t is None:
+                # TODO: Check if this should not be better to report as an error ?
+                self.configuration_warnings.append(
+                    "%s %r use/inherit from an unknown template (%r) ! Imported from: %s" % (
+                        type(item).__name__, item.name, name, item.imported_from                                                                                              )
+                )
+            else:
                 if t is item:
                     self.configuration_errors.append(
                         '%s %r use/inherits from itself ! Imported from: %s' % (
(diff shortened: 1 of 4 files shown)


gst:
This will fix problem mentionned in https://github.com/savoirfairelinux/monitoring-tools/pull/26

shinken/objects/item.py
```diff
@@ -998,8 +998,20 @@ def linkify_item_templates(self, item):
...
             t = self.find_tpl_by_name(name)
-            if t is not None:
-                tpls.append(t)
+            if t is None:
+                # TODO: Check if this should not be better to report as an error ?
+                self.configuration_warnings.append(
+                    "%s %r use/inherit from an unknown template (%r) ! Imported from: %s" % (
+                        type(item).__name__, item.name, name, item.imported_from                                                                                              )
+                )
+            else:
```
Seb-Solon: elif?

Seb-Solon:
Build broken, can't merge.

naparuba:
host calling an unknow template can be "normal" in the tag logic. But for
the others objects it's less true ^^

gst:
> host calling an unknow template can be "normal" in the tag logic. 
> But for the others objects it's less true ^^

that's sad ;)

Can you give an example where a Host (or any other (subclass-of-)Item eventually) would have a "use a_template_name"  and where the template "a_template_name" doesn't exist/isn't defined that would still be ok/valid for the host ? 

'cause I'm completly in favor of still having at least a configuration warning.
I'm in fact even more in favor of an configuration error, unless a specific case is found where it's valid to have an Host/Service/whatever have "use something" with the "somehting" not defined.. but I find that rather odd..

naparuba:
On Mon, Jan 19, 2015 at 8:45 PM, Grégory Starck <notifications@github.com>
wrote:

> host calling an unknow template can be "normal" in the tag logic.
> But for the others objects it's less true ^^
>
> that's sad ;)
>
> Can you give an example where a Host (or any other (subclass-of-)Item
> eventually) would have a "use a_template_name" and where the template
> "a_template_name" doesn't exist/isn't defined that would still be ok/valid
> for the host ?
>
that's the whole tag logic. User (not monitoring admins, real users like
standard admins) will set their own tags and pick some into the approved
tag from the first (that are doing packs).

With it you allow user to take the tag logic and then apply new packs if
you need it afterward. It's a user designed monitoring solution instead of
a expert based one. It's more genuine this way ^^

Some tag can be for example use for just... tagging :) like production or
thing like this, and just be search in UI or things like this.



>  'cause I'm completly in favor of still having at least a configuration
> warning.
> I'm in fact even more in favor of an configuration error, unless a
> specific case is found where it's valid to have an Host/Service/whatever
> have "use something" with the "somehting" not defined.. but I find that
> rather odd..
>
Nop, we should avoid to "break" configuration if we can. Of course a
warning or something like this when we thing there is a reason, but setting
too strong errors will make monitoring mangment only in the monitoring team
hand. And this is a very bad, bad idea if you want them to focus on the
pack instead of every day monitoring management (adding host or thing liek
this).

In a large env with numerous team, you will have errors in the
configurations from new users. Why stop monitoring and don't monitor 9999
server just because 1 is badly defined without a bad logic? (like a
parameter that don't affect other 9999 nodes, if the configuration is
"wrong" like a circular reference yes we should error it).

gst:
skip first part not really useful to me.

> Nop, we should avoid to "break" configuration if we can. Of course a
warning or something like this when we thing there is a reason, but setting
too strong errors will make monitoring mangment only in the monitoring team
hand. And this is a very bad, bad idea if you want them to focus on the
pack instead of every day monitoring management (adding host or thing liek
this).

Heuh... "Using/extending something ('use xxx') which is undefined" is simply (very/completly) bad to me ..  
if you want to "make use of xxx" then you should have xxx defined somewhere/somewhen.. oviously, shouldn't you ? :p 
or else then don't use xxx if it's undefined and you don't want to define it..

:s

naparuba:
No you are thinking as a dev in the conceptual part, or a monitoring
expert. But not as a simple admin user. That's not how they are seeing
configuration, and especially the tag logic. It should be harmess to "tag"
your hosts, it's all I said since year ^^

Yes from a dev or purely conceptual point of view it's strange. But users
are not dev :)

gst:
> No you are thinking as a dev in the conceptual part, or a monitoring
expert. But not as a simple admin user. That's not how they are seeing
configuration, and especially the tag logic. It should be harmess to "tag"
your hosts, it's all I said since year ^^

ok, well... I don't see/know, by heart, all the ins & outs that are possibly in effect here (Host/.. tags, and all that tag feature) so I will agree with you ;)

But the current warning is good so ?  if yes then let's merge :)

Seb-Solon:
We dont have any test to ensure regression guys. Please create one.

gst:
Test added.

naparuba:
Warning is ok as it don't stop the exec. Then it's a bit too much for hosts
but ok for all other types. But in some cases you want a template an
typoyed it. I never find a perfect balance between all of this. So a waring
can be great, as expert can still create all, and people that use tagging
logic won't look at logs I think ^^

gst:
wait before merge.. 

gonna add a test to get/check for the warning too..

gst:
here we are.. at your merge startup position ;)

PR #1471, opened 2015-01-21
Move "daemons Link" modules from shinken to shinken.objects

gst:
Still in the effort of consistency and better source tree,
and also as they are subclasses of Item (shinken.objects.item), 
and so to also stay logical (all Item subclasses should be defined in shinken/objects/): 

move *any_daemon*link.py files from shinken/ to shinken/objects/.

Done in a way back-compatible with the previous one and with a deprecation warning emitted when the modules are imported from the previous path.


gst:
@naparuba  ?

naparuba:
Good. I suppose the backward was for modules like livestatus?

It's better to have objects in /objects indeed :)

gst:
> Good. I suppose the backward was for modules like livestatus?

yes ; but not only for Livestatus but for any other (external to shinken) script/module that need to import shinken.xxxlink : there shouldn't be a lot but with the deprecation warning they will be noticed (if they enable deprecation warnings at all) about that change.

gst:
> but with the deprecation warning they will be noticed 

more important : they'll continue to work ;) (until we remove these back-compatible code/files)

PR #1473, opened 2015-01-26
Disable SSLv3 for CherryPy backend

colourmeamused:
Takes care of Poodle (https://www.poodlescan.com/) when SSL is used with cherrypy.
My instance was even allowing SSLv2, but that might be a Debian Wheezy issue.
Happy to clean / improve code further based on suggestions.

### DIFF
--- a/shinken/http_daemon.py
+++ b/shinken/http_daemon.py
@@ -42,6 +42,19 @@ try:
     from cherrypy import wsgiserver as cheery_wsgiserver
 except ImportError:
     cheery_wsgiserver = None
+try:
+    from OpenSSL import SSL
+    from cherrypy.wsgiserver.ssl_pyopenssl import pyOpenSSLAdapter
+    # Create 'safe' SSL adapter by disabling SSLv2/SSLv3 connections
+    class pyOpenSSLAdapterSafe(pyOpenSSLAdapter):
+        def get_context(self):
+            c = pyOpenSSLAdapter.get_context(self)
+            c.set_options(SSL.OP_NO_SSLv2 |
+                          SSL.OP_NO_SSLv3)
+            return c
+except ImportError:
+    SSL = None
+    pyOpenSSLAdapterSafe = None
 
 from wsgiref import simple_server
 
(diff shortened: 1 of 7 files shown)


gst:
only about the form :

I think OpenSSL isn't guaranteed to be available everywhere.

So if you can adapt the code so that if the import of OpenSSL fail then you fallback on current behavior ?
And same remark with the test : if OpenSSL is not importable -> skip test (using unittest.skipIf(.. ).

thx.

naparuba:
Great catch. Just that we should move the import in a try (it's better to exit with a real error than a importerror uncatched and raise by the daemon).

But when it's ok go for merging :)

colourmeamused:
Not a problem, I'll refactor OpenSSL and check why the Travis CI build failed and resubmit.

naparuba:
Cool thanks ^^

colourmeamused:
That should do it - sorry for code ugliness, still getting used to the Pythony way :-)

gst:
No code ugliness at all in all this :) you even cleaned some bad indentation and few little things.
and that looks now very good to me.

I'm +1 for having it merged anyway now.  
Just, if you know how to do it : if you could rebase it on nap/master and squash your 2 (non-merge) commits in one (the 2 other merge-commits will be cleaned out by the rebase on nap/master itself) ? then it'll be perfect :)

colourmeamused:
Commits all squashed up now!

Seb-Solon:
I don't think we should merge this right now. The commit include some pep8 of current code. This will create conflict with the big pep8 pull request. 

That's why having only one commit for a pull request is not a good idea when you are doing more that one thing. We should have like 1 commit for the fix, 1 for the pep8 and another one for the test (thanks for that by the way). It is really easy to read, and easy to understand.

shinken/http_daemon.py
```diff
@@ -3,7 +3,7 @@
 # -*- coding: utf-8 -*-
 
 # Copyright (C) 2009-2014:
-#     Gabes Jean, naparuba@gmail.com
+# Gabes Jean, naparuba@gmail.com
```
Seb-Solon: ?

gst:
> We should have like 1 commit for the fix, 1 for the pep8 and another one for the test (thanks for that by the way). It is really easy to read, and easy to understand.

well.. that would then be (one of) the most perfect PR ever submitted (to shinken ok) ;) 
more seriously: I think this is sufficiently in good form as it is. it's not a big change at all.. after all.

colourmeamused:
Part of it was a rather overenthusiastic IDE! I understand the point with the pep8 fix - I hadn't seen the other branch. Git merging with significant whitespace isn't fun.

naparuba:
Thanks )

pep8 is ok by the way.

PR #1475, opened 2015-01-26
Jenkins and python26

gst:
Mainly : made jenkins dedicated new_runtest.sh script now successfully pass the tests on python2.6 as well as on 2.7 and with "short" as well as "long" tests.

NB: there are still room for this script to be cleaned further but at least now it's (more) stable.

the other most important change is probably:
+ use "requirements.txt" files on shinken level, shinken/test, and shinken/test/jenkins.  ( instead of not common/usual "dependencies" or so ).
+ same for eventual python version specific requirements (requirements.py26.txt or requirements.py27.txt).

otherwise it's fix of some of the tests.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,2 @@
+pycurl
+importlib
--- a/shinken/dependencies
+++ b/shinken/dependencies
@@ -1,4 +1,0 @@
-python >= 2.6
-pycurl
-importlib
-mock  # only used for livestatus test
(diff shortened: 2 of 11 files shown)


gst:
@naparuba :
I propose that your merge first this PR : https://github.com/naparuba/shinken/pull/1476
then I'll rebase this one cause it will then conflict (just a little bit)..

thx.

gst:
there was a fail test run but it's an unstable one ; I've relaunched the test and it's ok.

whenever you want: you can so merge :)

PR #1478, opened 2015-01-29
Fix memory leak in default shinken file generation

t0xicCode:
The first file is never closed, which means that the memory is never freed (until the gc runs).


naparuba:
thanks, better with a real call indeed ^^

PR #1476, opened 2015-01-27
Made the code PEP8-compliant

ReAzem:
So yeah. We did some PEP8 tonight. All tests pass. We have also added a PEP8 check to Travis.

//cc @Seb-Solon @gst @titilambert @freddrickk  :beers: :beers: 

REBASES ARE COMING

PR #1480, opened 2015-01-29
Add new service_includes host directive

gst:
Same, but reversed, behavior than service_excludes directive.


gst:
@naparuba : Same remark than on https://github.com/naparuba/shinken/pull/1475 : 
please merge first the PEP8 PR first (https://github.com/naparuba/shinken/pull/1476) then I'll rebase this one if it would by then conflict.

naparuba:
you can rebase.

gst:
rebased.

but nb:  because of https://github.com/naparuba/shinken/issues/1481  
I'm now hesitating to do that work in this same branch .. just waiting the input from  @Seb-Solon to see..

gst:
fixed bad rebase

naparuba:
travis not hapy:

shinken/objects/service.py:1534:1: W293 blank line contains whitespace

gst:
> travis not hapy:
> shinken/objects/service.py:1534:1: W293 blank line contains whitespace

damn.. corrected.

Now that we (try to) follow PEP8 : 
I think we should better put pep8 check/execution before the tests run and not run tests if pep8 already fails..  we would not have to wait for all the tests to correctly finishes before see that, on its side, pep8 fails.. pep8 scan is far quicker than the tests run..

gst:
@Seb-Solon : what do you think of https://github.com/naparuba/shinken/pull/1480#issuecomment-72228439

naparuba:
I think about a warning at a begining, and a fail at the end (for quick
test). Because you launch tests during your code, but only pep8 at the end
before commit :)

gst:
For my part it's done.

Don't know if the 2 fix are enough.. or if there are others special cases that would be broken by the use of a service_excludes or service_includes directive.

gst:
Up to your merge !

shinken/objects/host.py
```diff
@@ -874,6 +881,31 @@ def add_host_chk_dependency(self, h, status, timeperiod, inherits_parent):
     def add_service_link(self, service):
         self.services.append(service)
 
+
+    def __repr__(self):
+        return '<Host host_name=%r name=%r use=%r />' % (
+            getattr(self, 'host_name', None),
+            getattr(self, 'name', None),
+            getattr(self, 'use', None))
```
Seb-Solon: This is not part of the pull request right? I mean, it's only to get fancy oupout not a technical need for the feature.

shinken/objects/host.py
```diff
@@ -874,6 +881,31 @@ def add_host_chk_dependency(self, h, status, timeperiod, inherits_parent):
...
+        ''' Check whether this host should have the passed service be "excluded" or "not included".
+
+        An host can define service_includes and/or service_excludes directive to either
+        white-list-only or black-list some services from itself.
+
+        :type svc: shinken.objects.service.Service
+        '''
+
+        sdescr = svc.service_description
+        if not svc.is_tpl() and self.service_includes:
```
Seb-Solon: Is it possible to call this function on non-template only? As geektphe reworked template and separated template for real object.

shinken/objects/item.py
```diff
@@ -904,37 +917,14 @@ def index_item(self, item, name=None):
...
         name_property = getattr(self.__class__, "name_property", None)
-        if name is None and name_property:
-            name = getattr(item, name_property, '')
-        try:
-            del self.name_to_item[name]
-        except KeyError:
-            pass
+        if name_property is None:
+            return
+        self.name_to_item.pop(getattr(item, name_property, ''), None)
```
Seb-Solon: pop return value which is not used here. why not use del?

shinken/objects/item.py
```diff
@@ -904,37 +917,14 @@ def index_item(self, item, name=None):
...
-
-        An optional `name` may be given to indicate the mame under which the
-        object was indexed.
-
+    def unindex_item(self, item):
+        """ Unindex an item from our name_to_item dict.
         :param item:    The item to unindex
-        :param name:    The name under which the item has been indexed.
         """
         name_property = getattr(self.__class__, "name_property", None)
```
Seb-Solon: Actually i'm figuring that there is no name_property for a  service object but there is one for an host one. Strange.

shinken/objects/item.py
```diff
@@ -878,24 +876,39 @@ def add_item(self, item, index=True):
...
+    def remove_item(self, item):
+        """Removes (and un-index) an item from our containers.
 
-        If an object holding the same name already exists in the index, the
-        conflict is managed by the `manage_conflict` method.
+        :param item: The item to be removed.
+        :type item:  Item  # or subclass of
+        """
+        self.unindex_item(item)
+        self.items.pop(item.id, None)
```
Seb-Solon: Same comment that L931-924

shinken/objects/service.py
```diff
@@ -543,6 +543,19 @@ class Service(SchedulingItem):
 #                        |___/
 ######
 
+    def __repr__(self):
+        return '<Service host_name=%r desc=%r name=%r use=%r />' % (
+            getattr(self, 'host_name', None),
+            getattr(self, 'service_description', None),
+            getattr(self, 'name', None),
+            getattr(self, 'use', None)
+        )
```
Seb-Solon: Same comment (styling only?)

shinken/objects/service.py
```diff
@@ -1227,6 +1241,8 @@ def get_downtime(self):
 
 # Class for list of services. It's mainly, mainly for configuration part
 class Services(Items):
+
+    name_property = 'unique_key'  # only used by (un)indexitem (via 'name_property')
```
Seb-Solon: oh, here you add them :)

shinken/objects/host.py
```diff
@@ -874,6 +881,31 @@ def add_host_chk_dependency(self, h, status, timeperiod, inherits_parent):
...
+
+    def __repr__(self):
+        return '<Host host_name=%r name=%r use=%r />' % (
+            getattr(self, 'host_name', None),
+            getattr(self, 'name', None),
+            getattr(self, 'use', None))
+
+    __str__ = __repr__
+
+    def is_excluded_for(self, svc):
```
Seb-Solon: Do you think you need a whole service object? You only use the service description and the is_tpl() (see my comment below). It would avoir the fake_service you are doing afterwards.

Seb-Solon:
See my comments ;)

shinken/objects/item.py
gst: because the key may be missing.

shinken/objects/item.py
gst: same answer ;)

shinken/objects/item.py
gst: there was no name_property ;)

shinken/objects/host.py
gst: I don't want to make a separate PR for this :p

shinken/objects/host.py
gst: don't know.. should have to check/see what geektophe reworked exactly..

shinken/objects/service.py
gst: greatly help for debug purpose, when you're in a debugger ;)
and this doesn't change anything hopefully (as we don't expect str_or_repr(a_service_or_a_host) to return something well defined).

So same answer : I won't make a separate PR for this :p

shinken/objects/host.py
gst: Damn you've spotted it.
I have to agree, grrrr ;-)

Seb-Solon:
Ok last thing to do before merge :  check if is_excluded_for call an see if the service is needed / we have a service template.

Seb-Solon:
Ok, we need the is.tpl() call because we use this function in explode_services_from_templates() which obviously uses template. 

So i think the better way here is to just add another parameter so that we have two variable (sdesc and is_tpl) annd we avoid the fake service creation 

Is this legit for you?

gst:
> Ok, we need the is.tpl() call because we use this function in explode_services_from_templates() which obviously uses template.
So i think the better way here is to just add another parameter so that we have two variable (sdesc and is_tpl) annd we avoid the fake service creation
Is this legit for you?

seems well yes, only have to give it a try..
thanks for the idea.

gst:
there..

nb: I'll rebase + squash once we agree..

Seb-Solon:
It's ok for me :) (I assume test will  pass)

Seb-Solon:
Missing doc actually. Fixing this

PR #1485, opened 2015-01-30
Clean: move Config.properties dict into its own/dedicated module [Proposed]

gst:
I think it's so long that moving it in its own module deserves it (and the code clarity of that dict itself and of the Config class).

I've done it in 2 possible ways, so that we can debate over that :p  

1) config_properties.py : just declare at this module level all properties as bare attributes/names of the module itself and then at the end build up the 'properties' dict itself (which must only contain Property instances in its values()) .

2) config_properties_dict.py : just have moved the dict from config.py to this module. nothing else.

visually I find both quite good ; though I've a small preference for 1 because I prefer "plain/bare" module attributes/names instead of directly string-keys in a dict. (also plain-name vs string-key eases IDE search (for smart search by symbol name)).

A sub-possibility is to change config.py as a (sub-)package and have config/properties.py  (config.py would be moved in config/\_\_init__.py).

Once we agree on which possibility I'll rebase and remove the other one.

and if we agree on this "clean" I propose to do this for all such (too) big dict declared/created :
- Config.types_creations,
- Host.properties and running_properties,
- Service.properties and running_properties
- are there others ?


naparuba:
I'm not for separate theses values from the code. It's easier to got both
definition and use in the same place. Code size is not a ral problem with
today ides with folding and such thing.

So I'm against the idea of separate data from code. (it's a data driven
code, so don't make sense to separate both ^^)

gst:
before directly closing can we debate a bit ? :D

> ..
> So I'm against the idea of separate data from code.

damn .. obviously separate data from code is *always* (even more when there are so much imo) "a good thing/practice to do" ..  I say *always* even if "*you should never say never nor should you never say always*" :|

gst:
> Code size is not a ral problem with today ides with folding and such thing.

code size imply code clarity.. and in this case I find it clearer to have that(such) big (not to say enormous) dict on/in its own module..

naparuba:
nop sorry. It's rude, but won't be flexible about this and I can't see how
I can accept the idea of separate data and code in a data driven code.

You can reopen if you want, but there is more chance I raise to 100KG in
the next year than I accept the idea of this :) so we can gain time and
swtich to another enhancement :)

naparuba:
Coe size is important, but got the whole thing about a subject is far more
important. Separate data and code in such a way code is working will make
the code harder to understand, far more than just a big file.

So no. But good try :)

gst:
it's **too** rude. frankly ;)

> Coe size is important, 

here we agree :)

> but got the whole thing about a subject is far more important. 

I don't see how you can see the *whole thing* entirely at all with such a big dict.. even in its own module it's still impossible to see it entirely.. so it's how could it be different from when it's directly created within Config ?

> Separate data and code in such a way

in such a way ? 
well, at least, the config_properties(_dict).py I proposed is in the same level than config.py itself .. :p

> code is working will make the code harder to understand, far more than just a big file.

or not..

gst:
> well, at least, the config_properties(_dict).py I proposed is in/at the same level than config.py itself .. :p

and in that perspective I've hesitated to do it by replacing the config.py as a subpackage, and have config/\_\_init__.py (with config.py content simply be moved here) and  config/properties.py to make it crystal clear/emphasize that this properties module is part of the config.

naparuba:
If it's in another file, it's in another part in your mind when you code.
But here it's strictly linked so must be in the same place.

It's not means to be rude, but allowing this will reduce comprehension for
the readers. So no.

gst:
> If it's in another file, it's in another part in your mind when you code.

:| When you work in a file you need to jump from it to possibly any other file as the need comes ..

> But here it's strictly linked so must be in the same place.

it's "linked" because it applies to the Config class ok, but I don't agree it means we have absolutely to have all that in the same file..

otherwise, going that way and if you stay totally logical then all shinken entirely should be in a single file.. which I guess you'll agree it totally bad and possibly impossible ;)

gst:
I reopen to have others point of view.. ;)

@Seb-Solon , @titilambert ,  hmm @ReAzem  ?
someone else ?

and just as bare info for all : this would make the config.py file goes from 2450 lines to roughly 1800 ; which is rather a good win in clarity regarding the only needed change (move 2 dicts) ; imho.

olivierHa:
I also prefer keeping everything in one big file for better code
readability with this file.

Splitting a file of 10000 lines split into 10 ten files of 1000 lines using
a logical view is fine and a good thing.

But here 2450 vs 1800 lines, it wasn't worth it ...

2015-01-30 19:06 GMT+01:00 Grégory Starck <notifications@github.com>:

gst:
welcome in the debate ;)

> I also prefer keeping everything in one big file for better code readability with this file.

I don't see the link with keeping a (already too) big file (imho) and code readability ;) but the opposite yes : a too big file impacts readability ; which is my whole point here ;)

> Splitting a file of 10000 lines split into 10 ten files of 1000 lines using a logical view is fine and a good thing.
> But here 2450 vs 1800 lines, it wasn't worth it ...

damn.. but 2450 to 1800 is already a ~31% win ! with only a (so simple) move of 2 dicts out of there !
if you agree that split 1x10000 to 10x1000 is a good thing then you should agree that 1x2450 to 1x~1800 + 2x~350 should also be a good thing too, shouldn't you ?

olivierHa:
sorry I mean "code comprehension" rather than "code readability" if you
prefer.

If you split a file of 20 lines into 2 files of 10lines, you got a 50% win
! Is your code more readable ? I doubt. Percentage means nothing here,
sorry. It is all about keeping code easy to read/understand.

2015-01-30 21:50 GMT+01:00 Grégory Starck <notifications@github.com>:

naparuba:
On Fri, Jan 30, 2015 at 10:33 PM, Olivier H <notifications@github.com>
wrote:

> sorry I mean "code comprehension" rather than "code readability" if you
> prefer.
>
> If you split a file of 20 lines into 2 files of 10lines, you got a 50% win
> ! Is your code more readable ? I doubt. Percentage means nothing here,
> sorry. It is all about keeping code easy to read/understand.
>
+1 :)

david-guenault:
@gst i'm not an expert like you but hey guy this is not an artistic project, beautiful code is not the subject ... i'm already lost in the number of files so if you add more i will give up. You say readability ? what the fuck about readability ?!!!??! what (new) contributors need is just a fast way to understand the code before contributing and that's not the way to do it ! Oh and a last thing, do this PR improve performance ? NO, do this PR add feature ? NO. If you read this you know what i think about that.

geektophe:
In my point of view, separating the code by putting the `properties` dictionary in another file won't make it much easier to read. Personally, I also prefer having the properties definition in the same place, it's much more easier to come back to it.

In addition, you'd break one of the coding standards: as @naparuba  said, Shinken is data driven. Each data object embeds its attributes definition set into it, and you know you may rely on it. If you start to split files, you'll quickly be lost between those that have their attributes exported and those that do not.

gst:
Ho boys.. ;) what disturbs for such a so simple and stupid move.. (which is quite the most simple possible refactor/split) whaaaoouuhh ;)  and whose goal was effectively only (shame) code readability (as often with refactor/splits), but apparently it is not that clear to everybody that it would make what it's/I intended it for. ( At least I've read that it would make it; to some point, easier to read ;^p) 

I'm closing otherwise I'm afraid to hear some that would be frightened to lose their job/wife/dog/cat or don't know what if we would do that (and i don't want some to lose their cat, it's non pythonic ;))

geektophe:
You're asking for people's opinion, don't you ? If it doesn't match yours, don't complain that much.

It's an open source project, and the reason why a majority disagreed are clearly exposed. Please respect other contributors mind.

david-guenault:
What i see is just a childish attitude, crying about a useless pr not being accepted. This is not correct to discredit @naparuba by reopening a pr closed and refused by the project leader and creator.

gst:
> This is not correct to discredit @naparuba by reopening a pr closed and refused by the project leader and creator.

WoWoW :| :| :|
I've reopen, absolutely not AT ALL to discredit anyone (I'm just astonished how a simple click on a button is being interpreted..), but to have others point of view, nothing more nothing else. And we've got some, though more negative than what I would have liked at the start.  Hey if I knew what reactions this would have resulted on from the start then I would not have reopen and asked for others thoughts about it (or not :p) .

Seb-Solon:
WTF? 

Guys, the Shinken community is looking ridiculous in this issue. Are you not old enough to be respectful to each other? If you cannot prevent yourself from being aggressive then ask someone else to write it for you. The pull request is about exhanging idea. If you feel offended in a some just say it in a private way. We dont need to open source this kind of behavior. 

Considering the pull request, readability depends on the point of view so there is no partial way to decide. As people diagree gst did the good thing to close it.

Regarding readability, IMO,  developer documentation will help a lot more to keep  / have new developper in Shinken.

Seb

PS : Chillout please, really.

david-guenault:
Please @Seb-Solon do not reverse the roles. The real point is that this started after @gst continue to ask about the pr even when @naparuba said no and said why it will not be accepted (and closed the pr). 

> If it's in another file, it's in another part in your mind when you code.
> But here it's strictly linked so must be in the same place.
>
> It's not means to be rude, but allowing this will reduce comprehension for
> the readers. So no.

naparuba:
I'll take some few hours before answering to all of this. A friend told me night can help to take good and more clear decisions. I'll apply this and I'll get back tomorrow about this.

naparuba:
Hi,

a good night is always great to better see things ^^

So here we got two things:
* misunderstood about what the code should be and which public is intended to hack it
* governance problem with a contributor that reopen a PR after I close it with a clear message about why it won't be merge.

As usual I'll try to find the root problem about it and not just say "oh this is bad and you should stop". Nope. We should get back to the misunderstood tree and solve it so it won't impact us in the future (see it as a root problem analysis for the framework but apply on the project level ^^ ).

In fact we are missing a clear governance text, about why, how we are doing this. I'm currently working on it and should be available in the incoming week after I look at it can solve nearly all issues we did have in the past.

This document will be crystal clear and will solve us such cases even before they happen (づ｡◕‿‿◕｡)づ


Ps: as with every time with such documents we can make everyone to understand but not everyone to accept it. But this will be managed in the document too ^^

titilambert:
@naparuba interesting :)
I'm really curious about which topics you will discussed in this document. Could you tell us more ? or at least, enumerate a few topics ? What do you mean "governance" ?

Anyway, good luck !

PS: just an advise, to be "crystal clear" you could try www.lojban.org

PR #1498, opened 2015-02-06
Enh: A failed action can miss its stderr output

gst:
which can be very usefull for debug/investigation ofcourse.

Usecase: the command exited with != 0 and only stderr output was given.

### DIFF
--- a/shinken/action.py
+++ b/shinken/action.py
@@ -209,21 +209,23 @@ class __Action(object):
         # we should not keep the process now
         del self.process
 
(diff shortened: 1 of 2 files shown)


Seb-Solon:
Looks like the test unclosing quote is failing because now it goes to your if and not the elif. :)

gst:
PEP8 OK :p

Seb-Solon:
Any tests that write on stderr and you can find in the output? :)

gst:
there

Seb-Solon:
Thanks!

PR #1500, opened 2015-02-06
Fix basemodule bad work() signature

gst:
Fix: BaseModule.work() signature not compatible with how Worker use and call it
    
nb: actually this method is only be used by "Satellite" daemons (poller, reactionner, receiver) for their external worker modules.

### DIFF
--- a/shinken/basemodule.py
+++ b/shinken/basemodule.py
@@ -87,10 +87,9 @@ class BaseModule(object):
     """
 
     def __init__(self, mod_conf):
-        """Instanciate a new module.
-        There can be many instance of the same type.
-        'mod_conf' is module configuration object
-        for this new module instance.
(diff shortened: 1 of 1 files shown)


Seb-Solon:
Receiver does not use it actually.

Do you have any test to add ? Like try to load a module and see that this is the new work method that is used to create the process.

gst:
> Do you have any test to add ? Like try to load a module and see that this is the new work method that is used to create the process.

The main method used by the worker processes has not changed (work()), only its signature, to match how it's actually called.

Seb-Solon:
I was more thinking about assert module.work something. Because I think previosuly a print  module.work was returning something like basemodule.main function 

The assert may be something where you have the target assignment I don't know if it's easy to exec this specific part of code without huge copy paste / run everything.

PR #1501, opened 2015-02-06
Enh http daemon overall throughput

gst:
Enh: (WSGI)http_daemon: don't create a fresh new thread for each http request
    
Actually use an hard-coded value as max number of connections/requests to process per client thread.
Higher value (eventually infinite) could also be used.
    
this doesn't change anything but prevent from having to spawn a new thread on each new http connection.
    
As there are quite a lot of http connections/requests between the shinken daemons this could eventually lower the overall system load.

### DIFF
--- a/shinken/daemon.py
+++ b/shinken/daemon.py
@@ -696,7 +696,7 @@ class Daemon(object):
         self.http_daemon = HTTPDaemon(self.host, self.port, http_backend,
                                       use_ssl, ca_cert, ssl_key,
                                       ssl_cert, ssl_conf.hard_ssl_name_check,
-                                      self.daemon_thread_pool_size)
+                                      self.daemon_thread_pool_size, self)
         # TODO: fix this "hack" :
         shinken.http_daemon.daemon_inst = self.http_daemon
 
--- a/shinken/http_daemon.py
+++ b/shinken/http_daemon.py
@@ -183,8 +183,10 @@ class WSGIREFAdapter (bottle.ServerAdapter):
 
 class WSGIREFBackend(object):
     def __init__(self, host, port, use_ssl, ca_cert, ssl_key,
-                 ssl_cert, hard_ssl_name_check, daemon_thread_pool_size):
+                 ssl_cert, hard_ssl_name_check, daemon_thread_pool_size,
+                 satellite_daemon):
         self.daemon_thread_pool_size = daemon_thread_pool_size
+        self.satellite_daemon = satellite_daemon
         try:
             self.srv = bottle.run(host=host, port=port,
                                   server=WSGIREFAdapter, quiet=True, use_ssl=use_ssl,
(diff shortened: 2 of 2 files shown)


Seb-Solon:
Tests? :D

gst:
> Tests? :D

they have all passed :p 
ok it's not a new feature although it could look like. More a "minor" change thus.

Seb-Solon:
Well it's hard to test really something. Maybe a more global benchmark about load / memory usage to show it's worth grouping http request in the same thread for a while.

gst:
it can only be better than previously.

Previously : 
http_daemon -> create N threads
each thread : handle_one_request -> exit
http_daemon -> if some threads exited -> create new threads to replace them

Now: 
http_daemon -> create N threads
each thread:  handle up to (currently)hardcoded number of requests/connections -> exit
http_daemon -> if some threads exit : create new threads to replace them

=> less threads to create (for the same number of incoming requests) : less work / more performance, better ;)

naparuba:
I think we should delete the whole wgi part, and only keep cherrypy one.
For example in SSL mode the keep alive is not accepted (http 1.0 only) and
so the CPU consumption is just TOOOOOOO heavy.

We can add cherrypy3 as a strong dep, after all I always say to people to
use it when they got problems, so will be better to force it, especially
cherrypy3 is easily available in the repos.

And this will also solve this case too ^^

Seb-Solon:
Sorry to reopen new topic but if you want to add a strong dep you will face the exact same problem we faced with importlib. It is easy to find the package but the default behavior is changed. So migration issue may appear and warning message should print etc.

gst:
I merely agree with Seb-Solon.. :x
and anyway :
>  It is easy to find the package 

doesn't mean it's necessarily also easy to have it installed/usable on any system setup..

gst:
> For example in SSL mode the keep alive is not accepted (http 1.0 only) 

hmmm.. I would need to check how things behave about that..

>and so the CPU consumption is just TOOOOOOO heavy.

if yes then that appears to me as a bug or at least a (very) bad behavior of this whole thing.. :s
but again I would need to check deeper to see the whole thing in action..

olivierHa:
It is a new "strong" dep, because every production shinken is already using
it.
Only test environment are maybe running without cherrypy.

We should put a warning in the changelog about it in the 2.4 release, and
it should be enough for this case no ?



2015-02-10 17:14 GMT+01:00 Grégory Starck <notifications@github.com>:

naparuba:
@Seb-Solon I was not ok to add a strong dep during the test phase few days before a release, not the same thing here. I was always clear that cherrypy was the only valid option for prod, and that wsgi will be deprecated in the future. Future is now :)

@gst for keepalive, it's by default in swsgi, with http 1.0. If you want to propose a patch to swsgi team you can ^^

So I'm against keeping the swsgi as it's a duplicated behavior, when a valid solution is possible and already installed on the production systems.

Seb-Solon:
>I was not ok to add a strong dep during the test phase few days before a release, not the same thing here.

Does this mean the importlib question can be raised again?

>I was always clear that cherrypy was the only valid option for prod, and that wsgi will be deprecated in the future. Future is now :)
Well the sane argument apply for imporlib / python 2.6. They will be deprecated. Concerning the only valid option I guess you mean between those two cause we hav plenty of http backend that are production ready IMO

>So I'm against keeping the swsgi as it's a duplicated behavior, 
So, in this case you are not for flexibility of the communication layer right? 

>when a valid solution is possible and already installed on the production systems.
Well same for imporlib IMO

PR #1502, opened 2015-02-07
Deletion of Windows folder

jfbutkiewicz:
No longer specific windows works. Not working with 2.X branch (specific packages for python, new config files... so the scripts are not working well and in most case break the standard installation) so let's fall with old type install with 2003 rk tools (srvany...) if needed on windows server and workstations. This will also update the language bar by removing a lots of C# codes. 
According with 10 min tutorial :
For 2.X+ the executable installer may not be provided. Consequently, installing Shinken on a Windows may be manual with setup.py. Steps are basically the same as on Linux (Python install etc.) but in windows environment it’s always a bit tricky...


Seb-Solon:
We were thinking about that last week :). Maybe we also should move the wmic rewrite to its own repository, shouldn't we?

jfbutkiewicz:
Of course, WMIC exe and config (in libexec-windows) and sources (in windows/tools/wmic) may also be moved if needed, but these programs are only work with windows. Without a real windows installation it does not make sense... but you may do as you want, does not matter for me !

titilambert:
+1 :)

PR #1504, opened 2015-02-09
tests: clean: Don't reset uselessly Scheduler.sched instance.

gst:
```
Scheduler(..)
```

is already instantiated within Shinken.\_\_init__() and linked to that Shinken' self instance: https://github.com/naparuba/shinken/blob/master/shinken/daemons/schedulerdaemon.py#L230

so I don't know why this was originally done but what I can say is that it looks awful :p 
( I'm quite astonished that this has not break some test, ouf ^^ )


naparuba:
Don't remember too :)

Tests are ok, let's merge.

PR #1507, opened 2015-02-10
Enh/Fix: restore Service.display_name previous behavior

gst:
That is: if a Service.display_name is not explicitly set/configured,
then return service_description.

display_name is obviously better to always be a str instance, nothing else.

This had side-effect (on livestatus already) since https://github.com/naparuba/shinken/pull/1480,
see for instance: https://travis-ci.org/shinken-monitoring/mod-livestatus/jobs/50215670

I'm assuming the tests will pass.. :o

### DIFF
--- a/shinken/objects/service.py
+++ b/shinken/objects/service.py
@@ -556,6 +556,17 @@ class Service(SchedulingItem):
     def unique_key(self):  # actually only used for (un)indexitem() via name_property..
         return (self.host_name, self.service_description)
 
+    @property
+    def display_name(self):
+        display_name = getattr(self, '_display_name', None)
+        if not display_name:
+            return self.service_description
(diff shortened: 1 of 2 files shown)


gst:
@naparuba or @Seb-Solon .. thanks to merge this one asap ;)

gst:
perfect.

olivierHa:
Hello, 

I am sorry, but this PR shouldn't have been merged. 
In shinken code, we use getter/setter not decorator. For code cohesion/comprehension, could you correct it ?

Otherwise, good catch !

Olivier

Seb-Solon:
It's the same thing actually https://docs.python.org/2/library/functions.html#property . Decorator are getter and setter

PR #1508, opened 2015-02-10
Fix: fixed unstable assert in test_continuous_notifications_delayed

gst:
which so failed from time to time.

This could be relatively easily reproduced by executing the test under high load,
which can be quickly done with a simple bash for loop :

from within shinken/test:

$ for x in $(seq 20) ; do ( nosetests -xv test_notifications:TestNotif.test_continuous_notifications_delayed &> res_$x || echo "===> $x failed ===" ) & done

and you nearly always got at least one failed test instance.

Now, starting up to 500 such test instances in parallel doesn't trigger anymore the failed assert :p

### DIFF
--- a/test/test_notifications.py
+++ b/test/test_notifications.py
@@ -26,6 +26,7 @@
 import time
 
 from shinken_test import unittest, ShinkenTest
+from shinken_test import time_hacker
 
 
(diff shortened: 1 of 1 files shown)


naparuba:
time race bugs are the worse ^^ gg

PR #1511, opened 2015-02-11
Enh: scheduler: don't create unnecessarily (many) copies of services and...

gst:
... hosts when you want to iter over them

particularly this one:

for elt in [y for y in [x for x in self.hosts] + [x for x in self.services]]:

which create a first list copy of hosts,
also another list copy of services,
then create a third one being the sum of the 2 previous ( [..] + [..] )
then create a fourth one [y for y in [..]+[..] ]
then iterate over the last one.

On big configs (more than thousands of services or hosts) this could actually quite help.

### DIFF
--- a/shinken/scheduler.py
+++ b/shinken/scheduler.py
@@ -146,7 +146,10 @@ class Scheduler(object):
                 self.broks, self.brokers:
             o.clear()
 
-
+    def iter_hosts_and_services(self):
+        for what in (self.hosts, self.services):
+            for elt in what:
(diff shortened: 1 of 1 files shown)


gst:
wait I've catched others places where the iter_hosts_and_services() can be also used..

gst:
done.

Seb-Solon:
It's ok for me !

naparuba:
Valid, thanks :)

Issue #1513, opened 2015-02-12
log formmater  in log


andyxning:
I think we should add each __human_timestamp_log__ to every daemon not just arbiterdaemon and schedulerdaemon. we should also make __naglog_result__ in log.py to accept a human readably time in its log not just a timestamp. It is very difficult to debug with timestamp. 

Time in below log item is very unreadable:
> [1423729784] ERROR: [Shinken] Ask actions to 0, got 5, [u'center0'], ['None']

I think we should only support time formatter like below:
> [%(asctime)s] %(levelname)s: %(message)s', '%a %b %d %H:%M:%S %Y'

not this one:
> [%(created)i] %(levelname)s: %(message)s

gst:
+1

andyxning:
I have make a PR about make the human readable timestamp log formatter the default and the only one. At the same time, i also change some test in order to make it work correctly. You can check it out in #1514

Seb-Solon:
Sorry guys, it's about Nagios compatibility. 

This format is made for log parsing. Tools need a timestamp to get event and so does modules. We wanted to remove human timestamp because it was a pain.

I think it written somewhere is docs that the logging format is designed for further processing.

You can still use the human format btw

andyxning:
if we can have the timestamp to follow the iso-8601 formatter, we can still
parse the timestamp easilly. Nothing more effort will be added, i think.



++++++
Ning Xie


2015-02-13 11:59 GMT+08:00 Sébastien Coavoux <notifications@github.com>:

naparuba:
We should keep nagios format for nagios logs, lot of tools are using it for reporting and such thing. We can't break this compatibility.

It's up to the modules to get logs and reformat it if need :)

Frescha:
> Am 13.02.2015 um 15:33 schrieb Gabès Jean <notifications@github.com>:
> 
> We should keep nagios format for nagios logs, lot of tools are using it for reporting and such thing. We can't break this compatibility.
> 
> It's up to the modules to get logs and reformat it if need :)
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/naparuba/shinken/issues/1513#issuecomment-74261867>.
> 
+1

andyxning:
@naparuba , can you tell me one of the tools that using the timestamp log formmater, so i can check it out.

Issue #1512, opened 2015-02-12
configuration error in scheduler-master.cfg


andyxning:
the module name of retention-redis should be __RedisRetention__ if you install it with __shinken install__.

Seb-Solon:
It is not up to dae on shinken.io. See https://github.com/shinken-monitoring/mod-retention-redis/blob/master/etc/modules/retention-redis.cfg

andyxning:
@Seb-Solon ， however the one downloaded from shinken.io throgh `shinken install retention-redis` will still use **RedisRetention**, may be you should re upload the newer one to shinken.io.

Seb-Solon:
Sorry, I cant because I'm not the first who updated this and I'm still not able to publish on all pack/module in shinken.io. @naparuba  could help on that.

PR #1514, opened 2015-02-13
fix human timestamp log and redis-retention module name in scheduler

andyxning:
fix about #1513 and #1512


Seb-Solon:
Log formatting should not be change IMO as discussed in #1513. Shinken internal also rely on that.

PR #1516, opened 2015-02-16
Change hashbangs to python2

arthru:
Using shinken on distributions where Python3 is the default was impossible due to Python2 only compatibility and incorrect hashbangs (``#!/usr/bin/env python``).

By explicitly telling python2 in hashbangs, Shinken still works on systems where there is only python2, and also compatible on systems where there is both Python2 and Python3.

This pull request change all hashbangs from ``#!/usr/bin/env python`` and ``#!/usr/bin/env python2.6`` to ``#!/usr/bin/env python2``

### DIFF
--- a/bin/shinken
+++ b/bin/shinken
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 # Copyright (C) 2009-2014:
 #    Gabes Jean, naparuba@gmail.com
--- a/bin/shinken-arbiter
+++ b/bin/shinken-arbiter
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 # Copyright (C) 2009-2014:
 #    Gabes Jean, naparuba@gmail.com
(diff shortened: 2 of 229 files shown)


titilambert:
Hello !
I have also one side question, why we keep hasbangs in non-executable python file ?
I think we can remove most of them, can't we ?

arthru:
Yes, hashbangs can be removed in non executable python file, but it is harmless to let them...

In that PR, I just used find/sed to bulk update them...

Shall I update my PR to remove hashbang from non executable files ?

gst:
Doesn't init script load at least one "default" file from /etc/default/ (yes, if one is present: SHINKEN_DEFAULT_FILE) ? 

if yes then you could put in it the overload of python to python2 in some way (have tried with an alias but doesn't work, an export of PATH with some dir prepended where python is a link to python2 would do it then) ?

arthru:
@gst : you won't launch ``shinken`` in an init script for example, so it is not as viable as setting python2 in hashbang.

arthru:
Any new advice on that ?

olivierHa:
I don't think it is a good idea to do that. python2 doesn't exist on all distros.
And the day, we will support python3, we will revert this PR.
It is up to the sysadmin to know that shinken is only working now with python 2.

hvad:
I think it's bad idea for "#!/usr/bin/env python2"

-1

PR #1522, opened 2015-02-23
Enh: converted last old style classes to newstyle

gst:
This help in many aspects. Already in debug.

But more important : this will ensure we have the behavior of newstyle classes (we want that).

This will also allow, in a near future, to move some (not to say most) of theses classes into shinken.objects as sublasses of Item, so to be fully consistent and be able to inherit common behavior from Item.

### DIFF
--- a/shinken/clients/livestatus.py
+++ b/shinken/clients/livestatus.py
@@ -25,10 +25,10 @@
 
 import socket
 import asyncore
-from log import logger
+from shinken.log import logger
 
 
-class LSSyncConnection:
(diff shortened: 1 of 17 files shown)


naparuba:
+1 ^^

shinken/clients/livestatus.py
```diff
@@ -25,10 +25,10 @@
 
 import socket
 import asyncore
-from log import logger
+from shinken.log import logger
```
Seb-Solon: Ninja modification spotted!

PR #1524, opened 2015-02-23
Factorize get_objects_from_from_queues() in shinken.daemon 

gst:
Function get_objects_from_from_queues() is duplicated (but with unnecessary differences between the duplicates).

-> factorized it in daemon.py.

Scheduler is a bit special.


naparuba:
Good clean :)

PR #1532, opened 2015-03-05
Fix module autogeneration and module attribute type

Seb-Solon:
Hi,

Need some review on that. Ping @titilambert @naparuba @gst @ others

This fixes two bugs : 
* The module named pipe is auto-generated before auto generating the arbiter. So if the arbiter was auto generated, the module was "forgotten" after. (no defined arbiter in conf but command_file statement in conf)

* Arbiter crash when trying to generate the named_pipe module (due to command_file statement). The type of the module attribute was wrong (list instead of str). 

The second test case must be enhanced A LOT so that we ensure type are correct for all variable.


titilambert:
It's good for me !

titilambert:
Folks (and me), please, make all your PR like this one (comments and tests), please :)

PR #1535, opened 2015-03-09
Fix raise error arbiterlink

Seb-Solon:
Hi there,

Need review on that, @titilambert @naparuba 

Case : The arbiter was not raising error found during parsing. There was an error introduced by the new indexing and it was not raised properly. This could lead to all other error in the arbiter not to be raised. Not easy to find problems. 

I removed the is_correct def because it was overloading the Items one and was doing bad. Items.is_config is working well, we don't need such overloading.


titilambert:
Thanks

Issue #1533, opened 2015-03-09
notification commands in contact and notificationway type Error


andyxning:
notification commands[[1](https://github.com/naparuba/shinken/blob/master/shinken/objects/contact.py#L64), [2](https://github.com/naparuba/shinken/blob/master/shinken/objects/contact.py#L65), [3](https://github.com/naparuba/shinken/blob/master/shinken/objects/contact.py#L79)] in contact and notificationway and notificationways in contact should be ListProp not StringPorp.

As the following processing about notification commands and notificationways are all treated as a list not a String, such as [1](https://github.com/naparuba/shinken/blob/master/shinken/objects/contact.py#L184), [2](https://github.com/naparuba/shinken/blob/master/shinken/objects/contact.py#L255).

The same problem exists in notificationway object.

At the same time, the doc in readthedoc([contact](http://shinken.readthedocs.org/en/latest/08_configobjects/contact.html, [notification](http://shinken.readthedocs.org/en/latest/08_configobjects/notificationway.html)) said that notification commands are list not a string.

Seb-Solon:
Thnks. I'm fixing this :)

Seb-Solon:
Actually i'm ok with the notificationways property but i'm not sure commands one should be a list. 

However : [this](https://github.com/naparuba/shinken/blob/9a3a34edaf7fdd58ba3b7164423203029df8421f/shinken/objects/contact.py#L166-166) makes me this that notificationsways property should be list and not a string. 

Then maybe it's legit to uniformize it.

Seb-Solon:
Ok [this(https://github.com/naparuba/shinken/blob/f34589fe6e66f4d6409dc27bf183c012ff4f737c/shinken/objects/schedulingitem.py#L1320-1320)] makes me this it should be the same for contact attribute. Working on it.

andyxning:
Thanks for the fix. And  yes, as far as i know, all the [_simple_way_parameters](https://github.com/naparuba/shinken/blob/2.4-RC1/shinken/objects/contact.py#L39) will be [explode](https://github.com/naparuba/shinken/blob/2.4-RC1/shinken/objects/contact.py#L278) into a notificationway bounded with the specific contact. and the actually used one is the notification as you [specified](https://github.com/naparuba/shinken/blob/master/shinken/objects/contact.py#L161).

PR #1536, opened 2015-03-10
Fix contact and notifway param type

Seb-Solon:
Hi, 

This fix the issue referenced in #1533  : type of notificationways were not the type defined in properties. This leads to side fix related to lists variable. 

Please proofread this PR and merge :)


naparuba:
Seems valid :)

Issue #1534, opened 2015-03-09
Problems with Macro Resolver since 2.2


5c077yP:
Hey I recently upgraded to shinken 2.2 and now some service macros are not resolved anymore. What I can see so far, it seems service macros are not inherited anymore.

The smallest config snippet would look like this:

```
# TEST HOST
define host {
  use                            my-host,generic-host ; usual generic-host
  host_name                      localhost
  address                        127.0.0.1
}

# ~~ pack definition

# HOST TEMPLATE
define host {
  name                            my-host
  register                        0
}

# SERVICE
define service {
  use                            my-service,generic-service ; usual generic-service
  service_description            CHECK-123
  host_name                      my-host
  check_command                  echo!$_SERVICEecho_command$
  # the check_command should be resolver to echo!OK | 'ok'=1" but it is not resolved at 
  register                       0
}

# SERVICE TEMPLATES
define service {
  name                            my-service
  _echo_command                   OK | 'ok'=1
  register                        0
}

```

This is just an example config that should show my use-case. It can see from the changelog that the macroresolver was fix from 2.0 => 2.2. I will try to look into the code and if I find something I will let you know, but maybe you already have an idea an can fix that simply before 2.4 final. 

Thanks in advance

5c077yP:
btw. this is the rest of my config (modules not included) if anyone wants to try

```
define host {
    name                            generic-host
    check_command                   echo
    check_period                    24x7
    contact_groups                  +admin
    register                        0
}

define service {
    name                            generic-service
    check_interval                  10
    retry_interval                  10
    check_period                    24x7
    contact_groups                  +admin
    register                        0
}

define command {
  command_name    echo
  command_line    /bin/echo '$ARG1$'
}

define timeperiod {
  timeperiod_name    24x7
  saturday           00:00-24:00
}

define contactgroup {
  contactgroup_name    admin
}

define contact {
  contact_name    me
  contactgroups   admin
}
```

5c077yP:
Hey, what i found out so far is:

- from 2.0 -> 2.2 `Items` was refactored to separate _items_ and _templates_  (into different dicts) from the beginning
- the `apply_inheritance` in `Items` as well as `Services` is __only__ called for the the _items_ but not _templates_ dict
- the `explode` step in arbeiterdaemon generates the items in my use-case (in 2.2)
- if I move `explode` before `apply_inheritance` (in arbeiterdaemon) it works, but i think this is __not__ the right way
[...]

PR #1537, opened 2015-03-11
fix tests so one can execute them from a clean repo

5c077yP:
Hey, 

i couldn't get the tests running locally when i just cloned the repo, installed dependencies (`pip install -r requirements.txt`) and run tests `./quick_test.sh` and `test.sh`. 

i wanted to contribute a PR for #1534 but i need this to get the tests running from the beginning

### DIFF
--- a/test/etc/test_sslv3_disabled/shinken.cfg
+++ b/test/etc/test_sslv3_disabled/shinken.cfg
@@ -38,7 +38,7 @@ cfg_dir=../core/realms
 #resource_file=resource.cfg
 
 # The path to the modules directory
-modules_dir=/var/lib/shinken/modules
+modules_dir=modules
 
 # Number of minutes between 2 retention save, here 1hour
 retention_update_interval=60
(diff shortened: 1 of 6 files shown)


test/etc/test_sslv3_disabled/shinken.cfg
```diff
@@ -114,7 +114,7 @@ local_log=/dev/null
 # By default don't launch even handlers during downtime. Put 0 to
 # get back the default N4G105 behavior
 no_event_handlers_during_downtimes=1
-
+modules_dir=modules
```
Seb-Solon: already line 41 actually

Seb-Solon:
Actually the test_sslv3 is disabled. the ssl.SSLContext is not available in python<=2.7.9.

Somehing else, I dont get why you change de import order. 

The RunTimeError fix is legit to me, I don't get why it was like this before

5c077yP:
concerning the import order. for me it was important to have the import shinken_test first as this is importing shinken properly (call to __import_shinken). otherwise the tests were failing and it told me that e.g. shinken has no module such as brok. 

i just cloned the repo , did the pip install -r requirements.txt and called to `quick_test.sh`

gst:
I'm also ok with this PR..

naparuba:
yes 2


thanks.

PR #1538, opened 2015-03-11
Feature/1534 fix item inheritance

5c077yP:
This PR fixes the handling of inheriting properties and custom variables. It is based on the issue #1534 and should be merged after #1537 

1. apply inheritance on both items and templates now
2. unify usage of apply_inheritance across the objects


Seb-Solon:
The fix is legit for me but it is over another PR actually. I'm not for merging the first one because tests need more refactoring .

I can reuse your commit to separate them properly if you don't mind. I'll do one commit for the fix and one for the test case related to the commit.

5c077yP:
i opened a new PR where only the the fix was cherry picked. see #1547

PR #1539, opened 2015-03-13
Fixed launch bin paths

ReAzem:


bin/launch_arbiter_debug.sh
```diff
@@ -23,7 +23,7 @@
 
 
 DIR="$(cd $(dirname "$0"); pwd)"
-BIN="$DIR"/../bin
+BIN="$DIR"/shinken/../bin
```
Seb-Solon: woops .

ReAzem: @Seb-Solon fixed~~

PR #1540, opened 2015-03-13
Fix setup.py for installation with zip file

titilambert:
Hello !
This is a small fix to be able to use ```python setup.py install``` outsite a git repository
To reproduce this problem:
```
git clone https://github.com/naparuba/shinken.git
cd shinken
rm -rf .git
python setup.py install
```

This patch fix it :)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,13 @@
 #!/usr/bin/python
+import os
+from shinken.bin import VERSION
+os.environ['PBR_VERSION'] = VERSION
+
 import setuptools
+
 
 setuptools.setup(
     setup_requires=['pbr'],
+    version=VERSION,
     pbr=True,
 )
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -1,7 +1,13 @@
 #!/usr/bin/python
+import os
+from shinken.bin import VERSION
```
naparuba: How can it import shinken if it's not already installed or in the PYTHONPATH?

Seb-Solon: Hum, in this case this I think python is doing a relative import . Setup.py is next to the shinken directory in the repo

titilambert:
This works for me. Could you confirm it on your env ?
Thanks !

gst:
> This works for me. Could you confirm it on your env ?

`python setup.py install` works for me too in this branch, modulo the pre-creation of a virtualenv.. (otherwise you missed sudo for a system install).

PR #1541, opened 2015-03-14
change article a to an, when in front of another vowel

kyzh:
http://en.wiktionary.org/wiki/a#Usage_notes_2


naparuba:
Thanks ^^

Issue #1543, opened 2015-03-20
bug in rc.d scripts for individual daemons


ryepup:
I'm looking to run just the poller daemon on one of my machines, installing shinken via pip. The daemons don't come back up after reboot as I'd expect.

I have a symlink in `/etc/rc2.d/S20shinken-poller` pointing at `/etc/init.d/shinken-poller`.

When I restart the box, I get no poller, and no new log entries.

When I run it manually, I get an error:

```console
$ /etc/rc2.d/S20shinken-poller start
etc/rc2.d/S20shinken-poller: 56: /etc/rc2.d/S20shinken-poller: /etc/rc2.d/shinken: not found
```

The problem is the `shinken-poller` init script expects to be run in the same directory as the `shinken` init script. It works fine if I run the `shinken-poller` script directly:

```console
$ /etc/init.d/shinken-poller start
Starting poller: 
   ...done.
```

The expected behavior is that `/etc/rc2.d/S20shinken-poller start` is equivalent to `/etc/init.d/shinken-poller start`

ryepup:
pull request coming shortly, this is an easy fix

Seb-Solon:
Thanks!

PR #1544, opened 2015-03-20
improve script directory finding in init scripts

ryepup:
the per-daemon init scripts delegate to the main `shinken` init
script. In order to do this, they find the current directory (where all
the init scripts are defined), and then calls `$curdir/shinken`.

The previous method of finding the current directory did not account for
symlinks, so running something like `/etc/rc2.d/S20shinken-poller
start` (where `S20shinken-poller` is a symlink)
would assume the current directory is `/etc/rc2.d/`, and then fail when
it couldn't find the main shinken init script.

This patch uses the method from http://stackoverflow.com/a/1638397 to
find the current directory of the script and resolve symlinks.

refs #1543

### DIFF
--- a/bin/init.d/shinken-arbiter
+++ b/bin/init.d/shinken-arbiter
@@ -45,8 +45,8 @@
 
 SHORTNAME=arbiter
 NAME="shinken-$SHORTNAME"
-
-curdir=$(dirname "$0")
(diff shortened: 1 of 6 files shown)


Seb-Solon:
Legit to me. I'll merge that once Travis is done (which succeed)

PR #1546, opened 2015-03-21
Try to describe packs and modules.

k0ste:
I do not know it will look better or not :)


naparuba:
Far better I think :)

thanks

PR #1547, opened 2015-03-23
Fix #1534

### LINKED ISSUE #1534: Problems with Macro Resolver since 2.2

5c077yP:
This PR will fix the inheritance of custom variables see #1534 and #1538 

cherry pick: "apply_inheritance on all items and templates. don't overwrite apply_inhe...ritance in subclasses where not necessary"

### DIFF
--- a/shinken/objects/config.py
+++ b/shinken/objects/config.py
@@ -1495,9 +1495,9 @@ class Config(Item):
         # print "Contacts"
         self.contacts.apply_inheritance()
         # print "Services"
-        self.services.apply_inheritance(self.hosts)
+        self.services.apply_inheritance()
         # print "Servicedependencies"
-        self.servicedependencies.apply_inheritance(self.hosts)
+        self.servicedependencies.apply_inheritance()
         # print "Hostdependencies"
         self.hostdependencies.apply_inheritance()
         # Also timeperiods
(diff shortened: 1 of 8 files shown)


Seb-Solon:
Thanks!

PR #1554, opened 2015-03-31
Fix #820

### LINKED ISSUE #820: Notificationmacro $NOTIFICATIONISESCALATED$ is always 'False'
I am using Shinken 1.2.4 and wanted to get some more information in my notification mails therefore added the macro $NOTIFICATIONISESCALATED$ to my notification command. I already did this with other macros in the same way so I don't think it is an configuration problem.

The notifications I now receive contain always 'False' for this macro even if the message is escalated and was sent to my mail address for escalations. 

Also the wiki says this macro should contain 0 or 1: http://www.shinken-monitoring.org/wiki/official/thebasics-macrolist

Seb-Solon:
This fix #820. The value for the macro ecalated is set at notification creation

### DIFF
--- a/shinken/objects/schedulingitem.py
+++ b/shinken/objects/schedulingitem.py
@@ -1289,7 +1289,7 @@ class SchedulingItem(Item):
     def scatter_notification(self, n):
         cls = self.__class__
         childnotifications = []
-
+        escalated = False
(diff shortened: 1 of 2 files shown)


titilambert:
Re thanks !

PR #1564, opened 2015-04-08
Enh: Also give check_command in broks.

gst:
EDIT:
This has been replaced by https://github.com/naparuba/shinken/pull/1592 but kept open for now..
See last comment.

-------
This allow the broker (more specifically one of its modules) to know what command was used for each check
 and eventually use that for its purpose.

### DIFF
--- a/shinken/commandcall.py
+++ b/shinken/commandcall.py
@@ -103,10 +103,6 @@ class CommandCall(DummyCommandCall):
         self.args = [s.replace('___PROTECT_EXCLAMATION___', '!')
                      for s in tab[1:]]
 
-    # If we didn't already lately relink us, do it
-    def late_linkify_with_command(self, commands):
-        pass
-
     def is_valid(self):
         return self.valid
 
--- a/shinken/commandcall.py
+++ b/shinken/commandcall.py
@@ -105,11 +105,7 @@ class CommandCall(DummyCommandCall):
 
     # If we didn't already lately relink us, do it
     def late_linkify_with_command(self, commands):
-        if self.late_relink_done:
-            return
-        self.late_relink_done = True
-        c = commands.find_by_name(self.command)
-        self.command = c
+        pass
 
     def is_valid(self):
         return self.valid
@@ -131,17 +127,6 @@ class CommandCall(DummyCommandCall):
         for prop in cls.properties:
             if hasattr(self, prop):
                 res[prop] = getattr(self, prop)
-
-        # The command is a bit special, we just put it's name
-        # or a '' if need
-        if self.command and not isinstance(self.command, basestring):
-            res['command'] = self.command.get_name()
-        # Maybe it's a repickle of a unpickle thing... (like with deepcopy). If so
-        # only take the value
-        elif self.command and isinstance(self.command, basestring):
-            res['command'] = self.command
-        else:
-            res['command'] = ''
 
         return res
 
--- a/shinken/objects/config.py
+++ b/shinken/objects/config.py
@@ -2221,8 +2221,11 @@ class Config(Item):
             # Now in packs we have the number of packs [h1, h2, etc]
             # equal to the number of schedulers.
             r.packs = packs
-        logger.info("Total number of hosts : %d",
-                    nb_elements_all_realms)
+        
+        for what in (self.contacts, self.hosts, self.services, self.commands):
+            logger.info("Number of %s : %d", type(what).__name__, len(what))
+
+        logger.info("Total number of hosts in all realms: %d", nb_elements_all_realms)
         if len(self.hosts) != nb_elements_all_realms:
             logger.warning("There are %d hosts defined, and %d hosts dispatched in the realms. "
                            "Some hosts have been ignored", len(self.hosts), nb_elements_all_realms)
--- a/shinken/objects/config.py
+++ b/shinken/objects/config.py
@@ -1258,23 +1258,6 @@ class Config(Item):
     def clean(self):
         self.services.clean()
 
-    # In the scheduler we need to relink the commandCall with
-    # the real commands
-    def late_linkify(self):
-        props = ['ocsp_command', 'ochp_command',
-                 'service_perfdata_command', 'host_perfdata_command',
-                 'global_host_event_handler', 'global_service_event_handler']
-        for prop in props:
-            cc = getattr(self, prop, None)
-            if cc:
-                cc.late_linkify_with_command(self.commands)
-
-        # But also other objects like hosts and services
-        self.hosts.late_linkify_h_by_commands(self.commands)
-        self.services.late_linkify_s_by_commands(self.commands)
-        self.contacts.late_linkify_c_by_commands(self.commands)
-
-
     # Some properties are dangerous to be send like that
     # like realms linked in hosts. Realms are too big to send (too linked)
     # We are also pre-serializing the confs so the sending phase will
--- a/shinken/objects/checkmodulation.py
+++ b/shinken/objects/checkmodulation.py
@@ -101,13 +101,6 @@ class CheckModulation(Item):
         return state
 
 
-    # In the scheduler we need to relink the commandCall with
-    # the real commands
-    def late_linkify_cw_by_commands(self, commands):
-        if self.check_command:
-            self.check_command.late_linkify_with_command(commands)
-
-
 class CheckModulations(Items):
     name_property = "checkmodulation_name"
     inner_class = CheckModulation
--- a/shinken/objects/contact.py
+++ b/shinken/objects/contact.py
@@ -251,12 +251,6 @@ class Contacts(Items):
             # Get the list, but first make elements uniq
             i.notificationways = list(set(new_notificationways))
 
-    def late_linkify_c_by_commands(self, commands):
-        for i in self:
-            for nw in i.notificationways:
-                nw.late_linkify_nw_by_commands(commands)
-
-
     # We look for contacts property in contacts and
     def explode(self, contactgroups, notificationways):
         # Contactgroups property need to be fullfill for got the informations
--- a/shinken/objects/host.py
+++ b/shinken/objects/host.py
@@ -1514,23 +1514,6 @@ class Hosts(Items):
                 for hg in h.hostgroups:
                     hostgroups.add_member(hname, hg.strip())
 
-
-    # In the scheduler we need to relink the commandCall with
-    # the real commands
-    def late_linkify_h_by_commands(self, commands):
-        props = ['check_command', 'event_handler', 'snapshot_command']
-        for h in self:
-            for prop in props:
-                cc = getattr(h, prop, None)
-                if cc:
-                    cc.late_linkify_with_command(commands)
-
-            # Ok also link checkmodulations
-            for cw in h.checkmodulations:
-                cw.late_linkify_cw_by_commands(commands)
-                print cw
-
-
     # Create dependencies:
     # Dependencies at the host level: host parent
     def apply_dependencies(self):
--- a/shinken/objects/notificationway.py
+++ b/shinken/objects/notificationway.py
@@ -232,15 +232,6 @@ class NotificationWay(Item):
         return state
 
 
-    # In the scheduler we need to relink the commandCall with
-    # the real commands
-    def late_linkify_nw_by_commands(self, commands):
-        props = ['service_notification_commands', 'host_notification_commands']
-        for prop in props:
-            for cc in getattr(self, prop, []):
-                cc.late_linkify_with_command(commands)
-
-
 class NotificationWays(Items):
     name_property = "notificationway_name"
     inner_class = NotificationWay
--- a/shinken/objects/service.py
+++ b/shinken/objects/service.py
@@ -1516,16 +1516,6 @@ class Services(Items):
                         s.configuration_errors.append(err)
             s.servicegroups = new_servicegroups
 
-    # In the scheduler we need to relink the commandCall with
-    # the real commands
-    def late_linkify_s_by_commands(self, commands):
-        props = ['check_command', 'event_handler', 'snapshot_command']
-        for s in self:
-            for prop in props:
-                cc = getattr(s, prop, None)
-                if cc:
-                    cc.late_linkify_with_command(commands)
-
     # Delete services by ids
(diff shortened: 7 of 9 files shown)


naparuba:
Valid and legit ^^

We can wait the rlease before merging, but can help lot of modules, like npcdmod for example, or even show real command launch in uis.

gst:
I m half happy of the result at broker level.. Don't mergeyet, gonna give it a second try..
Le 9 avr. 2015 3:05 AM, Gabès Jean <notifications@github.com> a écrit :Valid and legit ^^

We can wait the rlease before merging, but can help lot of modules, like npcdmod for example, or even show real command launch in uis.

gst:
continuing my thought (half happy) : 

in fact so at broker level we get this in the brok (after data has been unpickled)  : `brok.data['check_command'].command` which is actually the `command_name`..

checking that..

gst:
> in fact so at broker level we get this in the brok (after data has been unpickled) : brok.data['check_command'].command which is actually the command_name..

Ok I see the pattern : 

Each service and host have so its `check_command` property/attribute, and it is in fact a `CommandCall` instance (`shinken.commandcall` ; which doesn't inherit our Item and have a custom \_\_getstate\_\_)  ; a CommandCall instance has the `command` attribute which is a `Command` instance (`shinken.objects.command`) ; but all is in the \_\_getstate\_\_ of the CommandCall which, after normal loop over its properties has a special behavior, here it is: 

```
        # The command is a bit special, we just put it's name
        # or a '' if need
        if self.command and not isinstance(self.command, basestring):
            res['command'] = self.command.get_name()
        # Maybe it's a repickle of a unpickle thing... (like with deepcopy). If so
        # only take the value
        elif self.command and isinstance(self.command, basestring):
            res['command'] = self.command
        else:
            res['command'] = ''
```

where the command is so replaced/forced by the command_name (`res['command'] = self.command.get_name()`.. and also special cased for when pickle then unpickel then repickle of that commandcall happen.

Why not simply let the full Command instance be pickled/unpickled instead of replacing it with only the command_name (by the get_name() call) :-?

(this is also to be put in relation in regards to https://github.com/naparuba/shinken/pull/1522 ..)

gst:
Giving it a try..

gst:
Try travis-succeeded :)

this has permitted to simplify quite a bit the conf "relink" at scheduler level :)  because now the commandcalls always pickle their check_command instance. So no need to relink for each kind of commandcall anymore :)

I hope the ongoing travis-build is going to also succeed so .. ;)

gst:
NB: The error I want to know why it fails isn't related to this patchset / PR ..

gst:
@naparuba  @Seb-Solon 

I'm now rather happy with this..   quite a good simplification.. IMO : this makes more, with less.. ;)

I tried it a bit with my local dev shinken instance.. and couldn't see direct problems with that..

what do you think ?

naparuba:
If I remember the late linkify was to reduce the memory consumption or to recude pickle cost+size (for the arbiter).

If I'm not wrong it was a mix of both.

The idea was that commandcall where uniq between elements, but command objects must be uniq objects and relink.

So before merge i prefer we give a look on pickle size (arbiter->scehdulers and scheduler->broker) and in memory (as in memory command are a big part of the consumption).

gst:
> If I remember the late linkify was to reduce the memory consumption or to recude pickle cost+size (for the arbiter).

yeah I *was* sure about that also but now, as far as I see, this doesn't bring a problem apparently..

the pickle size of each CommandCall and Check will so now be (a bit) greater but appart from that I don't see any problem actually.

I'm gonna make more tests about this..

gst:
Using that config : 
```
[1429543071] INFO: [Shinken] Creating packs for realms
[1429543071] INFO: [Shinken] Number of hosts in the realm All: 504 (distributed in 504 linked packs)
[1429543071] INFO: [Shinken] Number of Contacts : 4
[1429543071] INFO: [Shinken] Number of Hosts : 504
[1429543071] INFO: [Shinken] Number of Services : 504
[1429543071] INFO: [Shinken] Number of Commands : 28
[1429543071] INFO: [Shinken] Total number of hosts in all realms: 504
[1429543071] INFO: [Shinken] Things look okay - No serious problems were detected during the pre-flight check
[1429543071] INFO: [Shinken] [Arbiter] Serializing the configurations...
```
(so not an "empty" config even if it could be larger)

I get, for master branch : 
```
[1429543189] DEBUG: [Shinken] [All] Serializing the configuration 0
[1429543189] DEBUG: [Shinken] [config] time to serialize the conf All:0 is 0.226522922516 (size:2386825)
[1429543189] DEBUG: [Shinken] PICKLE LEN : 2386825
[1429543189] DEBUG: [Shinken] [config] time to serialize the global conf : 0.194957971573 (size:4241853)
[1429543189] DEBUG: [Shinken] [config]serializing total: 0.421793937683
```

and for this branch I get : 

```
[1429543378] DEBUG: [Shinken] [All] Serializing the configuration 0
[1429543378] DEBUG: [Shinken] [config] time to serialize the conf All:0 is 0.204222202301 (size:2386599)
[1429543378] DEBUG: [Shinken] PICKLE LEN : 2386599
[1429543378] DEBUG: [Shinken] [config] time to serialize the global conf : 0.192235946655 (size:4241627)
[1429543378] DEBUG: [Shinken] [config]serializing total: 0.396622896194
```

which is about the same.. so no problem here we can say.


Now, at the scheduler to broker level : 

size of a host_check_result brok is ~1400 bytes in this branch against ~900-1000 in master branch..  yep that's a ~50% increase nearly.. :s 

This is because a Check instance is linked to a "CommandCall" instance, which is linked to the ("real") Command one... and so the CommandCall is also pickled as well as the Command, both have "quite a lot" (almost 10 for each) of attributes so the increase in the size of the serialization..

BUT: the broks when passed from the scheduler to the broker are compressed.. and the compression just makes that this increase isn't much of a problem finally (IMO), even more because there are many broks read by the broken in one (http-)get call .. -> many broks compressed -> better compression ratio -> less impact normally.

Well, I'm gonna give check to see if we couldn't lower that "relatively" easily ..

otherwise there would need something else..  like Commands definitions could be passed/given to each shinken daemon at arbiter start and then used "indirectly" by each of them instead of having  to (uselessly) pass their data in each brok..

gst:
> Well, I'm gonna give check to see if we couldn't lower that "relatively" easily ..

Half-succeeded "only" :  by "only" adding 'check_result' in *fill_brok* of Host and Service check_command attribute : 

we then also get the CommandCall instance of a particular check result at the broker level.. its 'command' attribute being so forced to the actual command_name (and not the Command instance) (by the specific CommandCall.\_\_getstate\_\_)..

So I get what I wanted (the command_name) but the brok size is still relatively higher : ~1200 bytes .. AND with that "half-"solution which doesn't permit to completely simplify all this pattern as it's done in this branch / patchset..  unfortunately ;) 

deep thoughts here..

gst:
Despite the increase of the (check_result-) broks bare size , but given that the broker fetch the broks many at a time AND compressed from the different daemons, so the increase is in fact, IMO, irrelevant : 

In `broker.get_new_broks()`, after / once the following code has executed : 
```
data = tmp_broks = con.get('get_broks', {'bname': self.name}, wait='long')
try:
    _t = base64.b64decode(tmp_broks)
    _t = zlib.decompress(_t)
    tmp_broks = cPickle.loads(_t)
```
at this point we have, here for ~7000 broks fetched (yes I had put a break point ;)) `len(data)` =~  800k but `sum(len(b.data) for b in tmp_broks.values())` =~ 4.5 M ; almost 6 times more.. so yes the actual size transferred between the daemons is irrelevant to the increase of size of a single brok (thanks to zlib)..

gst:
I have let my configuration (~500 hosts and ~500 services) run on the master branch and on this one.. during 15 minutes in each test run. 
And my system load was in both case between 0.6-7 and 0.8-9 depending on the time.. but average at 0.8 .. in both cases.. no real effective difference .. as far as I can conclude..

gst:
With a 1000 hosts and 4000 services configuration, after run of 15 minutes for each, (on my dev station but I had killed about every other service not related (so my browser with its 20+ tabs open too) to not interfere) : 

master :          load average: 0,84, 0,78, 0,74
this_branch:   load average: 0,88, 0,80, 0,76

I'm now more fairly sure that the change should impact in a noticeable way the performance of the whole solution.

wdyt ?

naparuba:
My main fear is about the boot time (pickle pass is what take us the most
time in the end). I'll give a test with a large env to see the overall
impact :)

gst:
For the same ~1000 hosts + ~4000 services configurations : 

this_branch : 
```
[1429622253] DEBUG: [Shinken] [config] time to serialize the conf All:0 is 1.22281694412 (size:12055417)
[1429622253] DEBUG: [Shinken] PICKLE LEN : 12055417
[1429622253] DEBUG: [Shinken] [config] time to serialize the global conf : 1.06914305687 (size:21123492)
[1429622253] DEBUG: [Shinken] [config]serializing total: 2.29213309288
```

master branch : 
```
[1429622352] DEBUG: [Shinken] [config] time to serialize the conf All:0 is 1.32023906708 (size:12055733)
[1429622352] DEBUG: [Shinken] PICKLE LEN : 12055733
[1429622352] DEBUG: [Shinken] [config] time to serialize the global conf : 1.3489010334 (size:21123808)
[1429622352] DEBUG: [Shinken] [config]serializing total: 2.6694149971
```

So there seems to even be some win with such a greater config than the first one I used :p

gst:
~5k hosts + ~20k  services : 

master : 
```
[1429622693] DEBUG: [Shinken] [config] time to serialize the conf All:0 is 5.93890213966 (size:60818044)
[1429622693] DEBUG: [Shinken] PICKLE LEN : 60818044
[1429622693] DEBUG: [Shinken] [config] time to serialize the global conf : 5.06226706505 (size:105835372)
[1429622693] DEBUG: [Shinken] [config]serializing total: 11.0014181137
```

this  : 
```
[1429622780] DEBUG: [Shinken] [config] time to serialize the conf All:0 is 5.81003379822 (size:60817637)
[1429622780] DEBUG: [Shinken] PICKLE LEN : 60817637
[1429622780] DEBUG: [Shinken] [config] time to serialize the global conf : 4.85905718803 (size:105834967)
[1429622780] DEBUG: [Shinken] [config]serializing total: 10.6693930626
```

That confirm quite clearly.. doesn't it ?

gst:
> My main fear is about the boot time (pickle pass is what take us the most
time in the end). I'll give a test with a large env to see the overall
impact :)

I'm more "affraid" by the eventual increase of processing required when running ;) but it seems we don't have to ..

Another "running" test, with 2500 hosts + 10k services, still 15 minutes run for each : 

this       >>  load average: 1,92, **1,68, 1,43**
master >>  load average: 1,61, **1,62, 1,47**

and the memory used by scheduler & broker stayed about the same between both branches:

this:
  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND
 5892 gstarck   20   0  887608 430572   2832 S  17,3  **5,3**   2:19.03 shinken-broker
 5639 gstarck   20   0 1056672 395008   3372 S  13,3  **4,9**   1:35.61 shinken-schedul

master:
17873 gstarck   20   0  880988 423924   2708 S  24,6  **5,2**   0:37.15 shinken-broker
17603 gstarck   20   0  993032 400276   3296 S  19,3  **4,9**   0:33.39 shinken-schedul

I'm quite impressed there doesn't have more differences ;)

I'm waiting your tests now..

shinken/objects/config.py
```diff
@@ -1258,23 +1258,6 @@ def remove_exclusions(self):
     def clean(self):
         self.services.clean()
 
-    # In the scheduler we need to relink the commandCall with
-    # the real commands
-    def late_linkify(self):
```
Seb-Solon: I don't get where this job is done now. I guess linkify still have to occur somehow right?

shinken/commandcall.py
```diff
@@ -132,17 +124,6 @@ def __getstate__(self):
             if hasattr(self, prop):
                 res[prop] = getattr(self, prop)
 
-        # The command is a bit special, we just put it's name
```
Seb-Solon: I guess here you remove the possibility the command is a string so that you don't need late linkify?

Seb-Solon:
I let you guyz benchmark this, I'm also OK with the PR if it does not fail on this point :)

gst:
the fact is that i did test this (i mean not only the travis tests which succeed but with a "full" shinken setup running locally), and so with lots (10k) of hosts & services (pointing to localhost) yes, but still a very very simple config (absolutely no "complex" rules, nor anything like that).. and I'm not completely sure about that actually.. ah I just spotted something (*.command.* related) to dig into.. coming back..

gst:
@naparuba :  don't know if you had time but if you made some test(s) on this then you should have also concluded that, from a performance point of view this doesn't impact (in a really noticeable way at least).. or not ?

gst:
I think I'm going to make a version which only add/assign the 'command_name' to a Check instance (and make it available through pickle ofcourse)....

and postpone this (if not cancel) 'cause despite it simplifies the "linking" phase at scheduler level (which is cool but that's all) the problem is that the "full" CommandCall and Command instances are always attached/pickled with each check_result_brok (and not sure if not also with other broks) which is in fact simply & completely overkill.. let's clearly say/recognize it.

gst:
> I think I'm going to make a version which only add/assign the 'command_name' to a Check instance (and make it available through pickle ofcourse)....

replace("Check instance").by("SchedulingItem check_result brok")

gst:
I have it, almost incredibly simpler.. coming in a moment in a new PR..

gst:
Ok so the first commit of this serie/patchset is now (better, at least for pickle size) handled by the PR https://github.com/naparuba/shinken/issues/1592 . 
So this first commit can simply be dropped.

BUT (yeah there is often one ;)) the 2 others commit could, in fact, still be desirable/possible I think.. 'because they still simplify the serialization(&deserialization) of the "global" config when passed from arbiter to scheduler(s) (or spare arbiter(s)).. 

So I leave this open for now..

gst:
closing it finally.

going to re-open another one with a better branch name : **enh_simplify_commandcall_serialization**

which is all about that here.

Issue #1562, opened 2015-04-07
Choose the right broker to push arbiter broks


andyxning:
Recently i have camed into a problem with two brokers, which one is master and the other is slave. However, the broks in  arbiterdaemon about all daemons in the HA environment will all be push to the slave master which has been in a status waiting for initial configuration and can do nothing about the received system(with WebUI's system tab) broks. All these broks should be passed to master broker when it is ok and only goes to slave broker when the master broker is down. 

related code is below(comments between "#“ lines are added by myself), all code can be found [here](https://github.com/naparuba/shinken/blob/master/shinken/daemons/arbiterdaemon.py#L213):
``` 
# We must push our broks to the broker
    # because it's stupid to make a crossing connection
    # so we find the broker responsible for our broks,
    # and we send it to him
    # TODO: better find the broker, here it can be dead?
    # or not the good one?
    def push_broks_to_broker(self):
        for brk in self.conf.brokers:
            # Send only if alive of course
            ######################################################
            # Here, we can not just use the **alive** attribute to determine which broker to use, 
            # because  slave brokers will also be **alive**. This is the mainly problem. 
           ##########################################################
            if brk.manage_arbiters and brk.alive:
                is_send = brk.push_broks(self.broks)
                if is_send:
                    # They are gone, we keep none!
                    self.broks.clear()
```

After read some source code about arbiterdaemon and dispatcher i think we can use the attribute **active** in satellitelink to be the marker of the "on-duty" broker. Since all the "on-duty" broker will have **active** be **True** in satellitelink.

andyxning:
I think i can fix this problem with a PR. Just wait for some time.

PR #1565, opened 2015-04-09
fix broker select error in arbiter

andyxning:
This pull request will fix the problem proposed in #1562, for more info please go to #1562。

### DIFF
--- a/shinken/daemons/arbiterdaemon.py
+++ b/shinken/daemons/arbiterdaemon.py
@@ -217,13 +217,24 @@ class Arbiter(Daemon):
     # TODO: better find the broker, here it can be dead?
     # or not the good one?
     def push_broks_to_broker(self):
-        for brk in self.conf.brokers:
-            # Send only if alive of course
-            if brk.manage_arbiters and brk.alive:
-                is_send = brk.push_broks(self.broks)
-                if is_send:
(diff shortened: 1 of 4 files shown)


naparuba:
Seems valid :)

I'll give a test.

Just  a pep8 error (look at https://travis-ci.org/naparuba/shinken/jobs/57749243 for it).

I'm waiting the 2.4 release before merge it, but when 2.4 is release and I test your code, will be merged :)

andyxning:
@naparuba , i will check the pop8 error and fix it soon.

naparuba:
Cool thanks :)

andyxning:
One more thing, i do not take care about whether the **push_broks** function will succeed. However, in the original implementation the **broks** should only be removed when **push_broks** will succeed. I have thought that there will be some problem if we continue to collect broks when all  the brokers down, thus we will continue to collect broks and consume the memory. It is dangerous only if we can make a threshold that we can not store more than **high_water_level** broks( or anything like that) .

If we just do not care about the execution status about ** push_broks** then we just have some status    broks lost and everything about schedule, check and notification will still work properly. When the downed broker daemon coming start again we can sent it with the newest status broks.

andyxning:
@naparuba , can you please take a look at this PR.

PR #1573, opened 2015-04-13
Fix: tests erroneously assumed that arbiterlink's id attribute would alw...

gst:
...ays be 0

which can be if the test script is run alone,

but so it can also be different if others tests were executed by the same (nose-)process instance, like with:

nosetests -xv test_action.py  test_arbiterlink_errors.py

find_by_name() is here for that.


Seb-Solon:
Legit to me

naparuba:
yes

naparuba:
(big lot of gif inside) ^^

PR #1575, opened 2015-04-14
test enh: slightly more robust for test_scheduler_init

gst:
should now always exit/terminate the scheduler process in every case.

could/should be TODO (but later): 

make the test runnable/passable when executed multiple times in //
(but it's the same for quite some others tests in shinken)


naparuba:
Cool, more stable tests ^^

PR #1579, opened 2015-04-21
Ensure strip function exist

lermit:
Hi,

When running the webui I get this error on host page 

  AttributeError: 'Contact' object has no attribute 'strip'
 
This pull request test the presence of the strip function before using it.

Regards

### DIFF
--- a/shinken/objects/contactgroup.py
+++ b/shinken/objects/contactgroup.py
@@ -52,7 +52,7 @@ class Contactgroup(Itemgroup):
 
     def get_contacts(self):
         if getattr(self, 'members', None) is not None:
-            return [m.strip() for m in self.members]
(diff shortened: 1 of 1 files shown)


Seb-Solon:
Hi,

Actually I don't think the webui should use this function. This function is used at parsing time when members are still not objects (string). For now I'll suggest to acces members or rebuild what you need because I think using dir is not a proper fix.

The underlying problem is that the type change between the beginning of parsing and runtime from str (like) to object. Maybe we should not strip at all (so we do not make an assumption on the list type) and strip earlier in the parsing phase.

lermit:
Hi,

Ok I understand. I will update the webui python script in order to use an other function.

Thank you very much for your concern.

Regard

PR #1583, opened 2015-04-23
Add support of proxy socks5 for shinken cli

ddurieux:


naparuba:
Thanks for the patch :)

PR #1586, opened 2015-04-23
Fix 1582

Seb-Solon:
This change the servicegroups field into a listprop so that + can be used.

Have fun reviewing it.


naparuba:
OK for me.

PR #1588, opened 2015-04-23
Fix launch check return value

### LINKED ISSUE #1587: bad assumption about launch_check() return value
spotted by .. reading the code :o

`def do_check_freshness(self)` [L183](https://github.com/naparuba/shinken/blob/master/shinken/objects/schedulingitem.py#L183)
returns either:
    `return self.launch_check(now)` [L204](https://github.com/naparuba/shinken/blob/master/shinken/objects/schedulingitem.py#L204)
or:
    `return None` [L211](https://github.com/naparuba/shinken/blob/master/shinken/objects/schedulingitem.py#L211)

`do_check_freshness` is (only) used in scheduler in [L1471](https://github.com/naparuba/shinken/blob/master/shinken/scheduler.py#L1471) : 
```
            c = elt.do_check_freshness()
            if c is not None:
                self.add(c)
```

`self.add` expects to get **some _object_** instance (like .. Check).. : 
```
    def add(self, elt):
        f = self.__add_actions.get(elt.__class__, None)
        if f:
            # print("found action for %s: %s" % (elt.__class__.__name__, f.__name__))
            f(self, elt)

    __add_actions = {
        Check:              add_Check,
        Brok:               add_Brok,
        Notification:       add_Notification,
        EventHandler:       add_EventHandler,
        Downtime:           add_Downtime,
        ContactDowntime:    add_ContactDowntime,
        Comment:            add_Comment,
        ExternalCommand:    add_ExternalCommand,
    }
```

BUT,  `launch_check` return None (if none check were launched, ok) or .. a check id ! not the Check instance itself, example [here](
https://github.com/naparuba/shinken/blob/master/shinken/objects/schedulingitem.py#L1368)
and [here](https://github.com/naparuba/shinken/blob/master/shinken/objects/schedulingitem.py#L1397)
and [here](https://github.com/naparuba/shinken/blob/master/shinken/objects/schedulingitem.py#L1454)

So Houston I think we got a problem ;)

gst:
This should fix #1587 

Waiting travis result..

### DIFF
--- a/shinken/external_command.py
+++ b/shinken/external_command.py
@@ -1551,14 +1551,11 @@ class ExternalCommandManager:
             if self.current_timestamp < host.last_chk:
                 return
 
-            i = host.launch_check(now, force=True)
-            c = None
-            for chk in host.checks_in_progress:
-                if chk.id == i:
-                    c = chk
+            c = host.launch_check(now, force=True)
             # Should not be possible to not find the check, but if so, don't crash
             if not c:
-                logger.error('Passive host check failed. Cannot find the check id %s', i)
+                logger.error('%s > Passive host check failed. None check launched !?',
+                             host.get_full_name())
                 return
             # Now we 'transform the check into a result'
             # So exit_status, output and status is eaten by the host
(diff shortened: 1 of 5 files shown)


gst:
damn... will look at that..

gst:
forgot an adaptation..

gst:
there it is.

obviously there is no test against the *feature* used by this piece of code : 
```
            c = elt.do_check_freshness()
            if c is not None:
                self.add(c)
```

otherwise it would(should) have detected that the check wasn't **_add_**ed ..

gst:
@naparuba  @Seb-Solon  :  up to you guys..

gst:
FYI, I'm adding a test case ..

gst:
there it is :) 

can I have my nice picture ? :p

Issue #1607, opened 2015-05-02
Catch trace error of a module of receiver


ddurieux:
When a broker module crash, we have the trace error of the module and so easy to fix it.
With receiver module, we don't have it. I have crash of nsca module, but not know what crash it

mohierf:
As an example : 
```
[1430636940] ERROR: [receiver-master] The external module nsca goes down unexpectedly!
[1430636940] INFO: [receiver-master] Setting the module nsca to restart
[1430636940] INFO: [receiver-master] Trying to init module: nsca
[1430636941] INFO: [receiver-master] I'm stopping module u'nsca' (pid=22519)
[1430636941] INFO: [receiver-master] Starting external process for instance nsca
[1430636941] INFO: [receiver-master] nsca is now started ; pid=22588
[1430636941] INFO: [receiver-master] [nsca[22588]]: Now running..
```

PR #1608, opened 2015-05-03
Catch error of module crash and run in external (multiprocessing). close...

ddurieux:
...s #1607


naparuba:
Legit, and tests are ok (automatic and manual).

"Just In time" pull request ^^

PR #1616, opened 2015-05-07
fix #1612

### LINKED ISSUE #1612: self.tags error in regenerator is wrong
it seems that the root reason is that the regenerator module has something wrong. Currently, i can tell that when we reload configuration, then the `self.tags` in `Regenerator` is wrong. It does not have anything to clean all previous tags count instead just adding.

            # Linkify tags
            for t in h.tags:
                if t not in self.tags:
                    self.tags[t] = 0
                self.tags[t] += 1
this will cause all the tags count number increase every time we reload the configuration.

This will also make WebUI to present some error info. 
and some related issues #1597

andyxning:
fix #1612, which will make self.tag accurate when used in shinken reloads.

### DIFF
--- a/shinken/misc/datamanager.py
+++ b/shinken/misc/datamanager.py
@@ -136,10 +136,18 @@ class DataManager(object):
     # Get the hosts tags sorted by names, and zero size in the end
     def get_host_tags_sorted(self):
         r = []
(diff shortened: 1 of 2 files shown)


andyxning:
@gst , can this PR be merged? I think we do take a look at this.

PR #1620, opened 2015-05-08
Add: a scheduler tick hook

gst:
So that shinken modules can use it to make some work on every scheduler tick.


naparuba:
Better to keep the same hook_name than in the other daemon: "tick" than a specific for scheduler no?

naparuba:
For example: https://github.com/gst/shinken/blob/scheduler_tick_hook/shinken/daemons/brokerdaemon.py#L786

PR #1644, opened 2015-05-28
Fix Default business rule notification options

Simage:
modify business rule notification options to be an empty list to prevent inadvertent supression of notifications


naparuba:
Thanks :)

PR #1645, opened 2015-05-28
add solaris SMF manifests

Simage:
Solaris SMF manifests to allow running on Solaris platforms under SMF


naparuba:
I don't know about solaris, but seems legit :)

PR #1646, opened 2015-05-29
Enh: Doc - Helpful resources added in Notifications article

melones:
New method of SMS notifications (Shinken SMS notifications using SMS hardware gateway SMSEagle) added in Docs. I thought it is a nice alternative for existing SMS notification methods.


geektophe:
Yes, I agree, it's definitely an useful resource. I noticed this equipment before. I never had time to test it, but it seems very interesting :)

The only thing I'm not sure about is if it's a good idea to include a link to a commercial product in the documentation of an open source project.

naparuba:
Thanks :)

naparuba:
Thanks

Issue #1113, opened 2014-02-25
Shinken sometimes doesn't close it's local log files when rotated


jccomputing:
On a regular basis, Shinken daemons don't close their local_log logfile when they are rotated. The logs are kept open and consume disk space, even if you can't list them.

This happens a lot with the broker, because it has several external modules that append in brokerd.log.

```
# lsof +L1
COMMAND     PID     USER   FD   TYPE DEVICE SIZE/OFF NLINK     NODE NAME
shinken-r 11712  shinken    5w   REG  254,2    13871     0    260123 /var/log/shinken/reactionnerd.log.2014-02-06 (deleted)
shinken-b 11798  shinken    5w   REG  254,2  2086037     0    260139 /var/log/shinken/brokerd.log.2014-02-06 (deleted)
shinken-b 11972  shinken    5w   REG  254,2    31191     0    260134 /var/log/shinken/brokerd.log.2014-02-24 (deleted)
shinken-b 11981  shinken    5w   REG  254,2  2086037     0    260139 /var/log/shinken/brokerd.log.2014-02-06 (deleted)
shinken-s 27407  shinken    6w   REG  254,2    46939     0    260115 /var/log/shinken/schedulerd.log.2014-02-17 (deleted)
```

Here is the tail of reactionner's stale log:

```
# debugfs /dev/mapper/mainvg-varlog -R 'cat <260123>' | tail
2014-02-06 15:30:18,248 [1391697018] Warning : [reactionner-master] Scheduler scheduler-master is not initialized or has network problem: unknown object
2014-02-06 15:30:18,248 [1391697018] Warning : Sent failed!
2014-02-06 15:30:18,537 [1391697018] Warning : Printing stored debug messages prior to our daemonization
2014-02-06 15:30:51,246 [1391697051] Warning : [reactionner-master] Scheduler scheduler-master is not initialized or has network problem: receiving: timeout
2014-02-06 15:30:57,641 [1391697057] Warning : [reactionner-master] Scheduler scheduler-master is not initialized or has network problem: receiving: timeout
2014-02-06 15:30:58,195 [1391697058] Warning : Sent failed!
2014-02-06 16:29:51,129 [1391700591] Warning : [0] Timeout, Harakiri
2014-02-06 16:29:51,889 [1391700591] Warning : [reactionner-master] The worker 0 goes down unexpectedly!
2014-02-06 23:29:52,192 [1391725792] Warning : [1] Timeout, Harakiri
[...]

PR #1649, opened 2015-05-31
Handle log_rotation_method none, hourly and weekly

cedef:
Hi,

Here is a small patch I would like to submit in order to workaround bug #1113 
Basically I'd like to use log_rotation_method=n in order to let logrotate do the job using either a *postrotate script* or a *copytruncate* 

By the way, I wonder why the log_rotation_method should only be considered for *broker server* (as written in the documentation) rather than for all daemons ? (To me it currently applies to all daemons, but I might have missed something ?)

Please, let me know if anything needs to be improved before merging.

Regards,

### DIFF
--- a/doc/source/03_configuration/configmain-advanced.rst
+++ b/doc/source/03_configuration/configmain-advanced.rst
@@ -785,7 +785,7 @@ This is the rotation method that you would like Shinken to use for your log file
   * w = Weekly (rotate the log at midnight on Saturday)
   * m = Monthly (rotate the log at midnight on the last day of the month)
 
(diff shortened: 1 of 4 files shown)


cedef:
Ahem, 

There is something fishy .... I've made a stupid mistake on this one.
I've not tested enought my code and only arbiter daemon has a **self.conf** object.

By the way, it now looks strange to me to have the *log_rotation_method* parameter inside *shinken.cfg*... Should it go inside *daemons/*.ini* files ?

Anyway, I would be happy to implement this feature but I will need an advice on how/where to do it the right way.

Regards,

PR #1650, opened 2015-05-31
Create check_pop.cfg command

0pc0deFR:
In commands folder, the check_pop.cfg isn't present.


naparuba:
Such command should be in a pack instead ofthe framework :) 

But you can enhance the check_pop definition with your helful command on the pop pack :+1: 

https://github.com/shinken-monitoring/pack-pop3/tree/master/pack

Issue #1088, opened 2014-02-13
shinken2.0 cannot start webui


itxx00:
after installed shinken 2.0 and webui and restarted all the services,
the webui cannot startup, and no program listen on port 7767.

https://asciinema.org/a/7604

Hlearner:
Pleaaaase i've installed shinken 2.4 with pip method, ans add the webui module declaration to the broker but still the webui can't start, the broker log has no message error and i don't know where is the problem, i've installed shinken many times but i always get the same problem
i am on vmware ubuntu 14.04 
could anyone help me pleaaaaaaaaaaaaase

fninja:
same here.
i followed the steps as specified above.
broker log gives me this:
```
[1461592545] WARNING: [broker-master] The module type webui for WebUI was not found in modules!
[1461592545] INFO: [broker-master] I correctly loaded the modules: []
```
No webui listening on `:7767` :
```
$ sudo netstat -plnt | grep 776*
tcp        0      0 0.0.0.0:7768            0.0.0.0:*               LISTEN      51247/python2.7 
tcp        0      0 0.0.0.0:7769            0.0.0.0:*               LISTEN      51288/python2.7 
tcp        0      0 127.0.0.1:7770          0.0.0.0:*               LISTEN      51350/python2.7 
tcp        0      0 0.0.0.0:7771            0.0.0.0:*               LISTEN      51266/python2.7 
tcp        0      0 0.0.0.0:7772            0.0.0.0:*               LISTEN      51312/python2.7 
tcp        0      0 0.0.0.0:7773            0.0.0.0:*               LISTEN      51333/python2.7
```

PR #1651, opened 2015-05-31
Add note about adding webui to module in broker

MorrisJobke:
It costs me a lot of time to find out, why the web UI didn't started. Found the solution in naparuba/shinken#1088 ... but this was a really bad first experience that should be avoided ;)


naparuba:
Yes ^^

Thanks for the patch ;)

PR #1655, opened 2015-06-05
Import time in livestatus client even if client is not used directly as a script.

drakaz:
We import livestatus client in a trigger in a multi-instances environment, so import time is needed for both usage (script/import).


naparuba:
Valid, thanks :)

PR #1658, opened 2015-06-11
Correct hostescalation example

s-dewailly:
first_notification is the number of notification sent, and here we can
see a time « 60 ».


naparuba:
True, thanks for sharing :)

PR #1666, opened 2015-06-22
add hint to webui module name

xorpaul:
Otherwise the user might simply add 'webui' to the broker-master.cfg and encounter an error on the next shinken restart.

```
[item::broker-master] Error: the module webui is unknown for broker-master
```


xorpaul:
Okay after figuring out, that the webui module was still missing from my system (I used pip install shinken) and following http://shinkenlab.io/online-course-2-webui/ it worked with module webui in broker-master.cfg.

So you can ignore this pull request.

PR #1674, opened 2015-07-10
Update service.rst

nslythe:
Ther wa an errors in the duplicate_foreach section.
We have to put a "$" at the end of the line and there is missing "," between each KEY


olivierHa:
Thanks !

PR #1675, opened 2015-07-11
Downtimes broks

mohierf:
Added a brok when an host downtime is scheduled
Added a brok when a service downtime is scheduled

Allows to develop broker modules that are aware of those events ... like for sending notifications to an helpdesk ticketing system.


naparuba:
Thanks :)

mohierf:
Welcome home :-)

PR #1653, opened 2015-06-03
Shinken default notification plugin

claneys:
Shinken email notification plugin.

It send notification in a txt or html format, including url to webui if wanted and add a logo at the top to include by example customer company logo.

PR #1680, opened 2015-07-20
Shinken 2.4.1 html dependency

k0ste:
New html notifications required [this](https://pypi.python.org/pypi/html).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,3 +2,4 @@ CherryPy
 pycurl
 importlib
 pbr
+html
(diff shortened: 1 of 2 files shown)


hvad:
Hi,

I don't think HTML Notification is necessary for monitoring tool. 

Best regard

k0ste:
It's already required by #1653 because is default for Shinken 2.4.1

claneys:
My bad, forgot that requirement. Thanks you !

hvad:
Hi don"t agree with this commit because HTML notification is not good idea and non sense with webui2.

k0ste:
@hvad, for this time Shinken user can't make choice, html or not #1653.

Issue #603, opened 2012-09-12
inheritance of dependencies from templates


tvtue:
More info about this in the forum:
http://www.shinken-monitoring.org/forum/index.php?topic=545

Seb-Solon : Copy here forum is moving
# 

Hi,

I am trying to configure service dependencies for all nrpe checks of a host to the corresponding nrpe daemon on the same host. The problem is to write an elegant and short config for this.

Say I have 5 servers: server01, server02 ... server05. On each server the nrpe checks load, mem and disk.

I started creating servicedependency definitions for each nrpe check like so:

```
define service {
   use      my_service, nrpe
   hostgroup_name   my_servers
}

define service {
   use      my_service, load
   hostgroup_name   my-servers
}

define servicedependency {
   host_name            server01
   service_description   nrpe
   dependent_host_name   server01
   dependent_service_description   load
   notification_failure_criteria   c,u
   execution_failure_criteria   n
}

define servicedependency {
   host_name            server02
   service_description   nrpe
   dependent_host_name   server02
   dependent_service_description   load
   notification_failure_criteria   c,u
   execution_failure_criteria   n
}

define servicedependency {
   host_name            server03
   service_description   nrpe
   dependent_host_name   server03
   dependent_service_description   load
   notification_failure_criteria   c,u
   execution_failure_criteria   n
}
...
```

but that is lame and cumbersome.

Then I tried to use a servicegroup in the servicedependency definition:

```
define servicedependency {
   hostgroup_name           my_servers
   service_description   nrpe
   dependent_hostgroup_name   my_servers
   dependent_service_description   load, mem, disk
   notification_failure_criteria   c,u
   execution_failure_criteria   n
}
```

This "connects" every nrpe on any host with any load, mem and disk on any host. Kind a like an cartesian product. But that is wrong. What I need is a one on one dependency.

Then I asked in the monitoring portal for an elegant solution. Someone came up with the "same host depedency" trick, which I find to be very good. It can be achieved by leaving out the "dependent_host_name" or "dependent_hostgroup_name" line in the definition.

I tested this with only two hosts and one service on my Shinken 1.2 installation and with the following config:

```
define servicedependency {
   host_name            server01, server02
   service_description   nrpe
   #dependent_host_name   <empty> for same host dependency
   dependent_service_description   load
   notification_failure_criteria   w,c,u
   execution_failure_criteria   w,c,u
}
```

According to the "Impact Graph" Shinken still connects any nrpe services with any load service on every host. 

[...]

PR #1683, opened 2015-07-20
Svcdependencies auto computing

bougetq:
This pull request is motivated by the fact that servicedependencies are rather useful but really complicated to correctly setup right now, especially on large clusters. Currently the 'host_name', 'dependent_host_name', 'hostgroup_name'  and 'dependent_hostgroup_name' properties do not allow users to define host by host servicedependencies or service to service dependencies at once.
The idea is to compute the hosts that have the dependent service and the service depended upon and create a servicedependecy for each one of them.
For more details please look at the message from 3rd commit.


bougetq:
I also fixed the host to host servicedependencies creation as described [here] (http://shinken.readthedocs.org/en/latest/07_advanced/objecttricks.html#advanced-objecttricks-same-host-dependencies) and reported as an issue : #603 
I would like instructions as to how I am supposed to document this patch if you are interested in it.

naparuba:
Thanks and sorry for the delay. I4l have a close look for 2.6 (I'm getting PR for 2.4.2 currently)

bougetq:
Thanks for the reply. When you get the chance, please also take a look at [this] (https://github.com/bougetq/shinken/tree/nodesetparse) as both have been integrated at the CEA and are working just fine.

olivierHa:
This seems legit and a new great feature ! @naparuba , any suggestions ? Can I merge it ?

PR #1690, opened 2015-07-31
Enh: Add id getter for objects

Seb-Solon:
Like notification, check and event handlers object. 

Id should be private, it's a better way to access those field for external modules for example.

### DIFF
--- a/shinken/comment.py
+++ b/shinken/comment.py
@@ -74,6 +74,14 @@ class Comment:
     def __str__(self):
         return "Comment id=%d %s" % (self.id, self.comment)
 
+    def get_id(self):
+        """Getter for id attribute
+
+        :return: id
+        :rtype: int
+        """
+        return self.id
+
     # Call by pickle for dataify the ackn
     # because we DO NOT WANT REF in this pickleisation!
     def __getstate__(self):
--- a/shinken/downtime.py
+++ b/shinken/downtime.py
@@ -113,6 +113,14 @@ class Downtime:
         return "%s %s Downtime id=%d %s - %s" % (
             active, type, self.id, time.ctime(self.start_time), time.ctime(self.end_time))
 
+    def get_id(self):
+        """Getter for id attribute
+
+        :return: id
+        :rtype: int
+        """
+        return self.id
+
     def trigger_me(self, other_downtime):
         self.activate_me.append(other_downtime)
 
--- a/shinken/objects/item.py
+++ b/shinken/objects/item.py
@@ -211,6 +211,14 @@ Like temporary attributes such as "imported_from", etc.. """
     def __str__(self):
         return str(self.__dict__) + '\n'
 
+    def get_id(self):
+        """Getter for id attribute
+
+        :return: id
+        :rtype: int
+        """
+        return self.id
+
     def is_tpl(self):
         """ Return if the elements is a template """
         return not getattr(self, "register", True)


naparuba:
You push on the wrong repo.

ddurieux:
It's a function used by new version of livestatus module, so need to have it in Shinken ;)

naparuba:
? But what is the goal of a one line getter unless make the code less clear?

naparuba:
ps: module should be patch after the core, not the other way. Please ask before pushing to modules for such things.

naparuba:
question: what is the behavoir of the module with older core version?

naparuba:
edit: ok so on another livestatus repo, I though it was on the master one. so the order is still ok :)

I'm nto a big fan of getter/setter, especially with just one line. so what is the real reason of theses functions?

naparuba:
(please avoid to say it's for clarity or such things, you know that it make the code harder to understand).

olivierHa:
Old debate : 

http://www.javaworld.com/article/2073723/core-java/why-getter-and-setter-methods-are-evil.html
http://www.yegor256.com/2014/09/16/getters-and-setters-are-evil.html

And I quote :  http://dirtsimple.org/2004/12/python-is-not-java.html

Getters and setters are evil. Evil, evil, I say! Python objects are not Java beans. Do not write getters and setters. This is what the 'property' built-in is for. And do not take that to mean that you should write getters and setters, and then wrap them in 'property'. That means that until you prove that you need anything more than a simple attribute access, don't write getters and setters. They are a waste of CPU time, but more important, they are a waste of programmer time. Not just for the people writing the code and tests, but for the people who have to read and understand them as well.

Seb-Solon:
Hi,

I will be as clear as possible : 

* Such getters already exists in the following classes : 
    Notification : https://github.com/naparuba/shinken/blob/master/shinken/notification.py#L183
    Check : https://github.com/naparuba/shinken/blob/master/shinken/check.py#L140
    Event_handlers : https://github.com/naparuba/shinken/blob/master/shinken/eventhandler.py#L122
You can git blame to see authors guys, I did not write those lines.
The goal here is to add consistency and let other object have such getters. If you think those line are bad why it's still there since 2009?
Getter / Setter / Property is not really a problem, I just followed the code logic already here.

* id is a reserved keyword for python : https://docs.python.org/2/library/functions.html#id. Overriding it is bad : http://pylint-messages.wikidot.com/messages:w0141

* id should be private or at least not used by external modules (like livestatus). By doing this you remove an implementation dependency between core and module. : https://en.wikipedia.org/wiki/Loose_coupling . 

* I did not to pushed to the wrong repo, just trying to keep on contributing in Shinken. It's Open source right?

* If someone is able to understand the get id for the previous classes I assume he will be able to understand get id for those ones. So there is no readability debate.


I'm trying to make livestatus working with both cores which is basically the free software state of mind.

gst:
For info: I don't see also the actual gain/win of this..

I +1 Olivier comment: https://github.com/naparuba/shinken/pull/1690#issuecomment-127230712

Pythonic way to implement getter & setter in python, **when need to be**, is with property. Otherwise go directly with the bare/direct attribute access.

> id is a reserved keyword for python : https://docs.python.org/2/library/functions.html#id. Overriding it is bad : http://pylint-messages.wikidot.com/all-messages

it's far far far (but really) far less bad when used for an object attribute.. It's when it's used as a bare **name** that it's more problematic, either within a module (the more problematic) or within a function (the less). As an object attribute it's even less problematic so.

NB: IMO : it's not that it's that "bad" to override/rename python builtins names, but it's that it can be problematic (if/when you want to use the original/builtin value) and can lead to obvious bugs.

> id should be private or at least not used by external modules (like livestatus). By doing this you remove an implementation dependency between core and module. :

If it is private, then why should it be used by external modules (because that the purpose of this commit) :-?

PR #1703, opened 2015-08-25
Fix: service_(includes|excludes) tmpl recursion

geektophe:
This patch Fixes a regression introduced in
9f9bc2040289bd897a9064c6445ebc08221440f9

If an attribute covered by a property is explicitly set at class or instance level (during initialization for example), it breaks attribute's template recursion in `get_property_by_inheritance` on
first `if hasattr(self, prop)` instruction (attributes is always seen as defined, even if it was never set).

In the present case, it broke service exclusion defined on templates rather than host.

### DIFF
--- a/shinken/objects/host.py
+++ b/shinken/objects/host.py
@@ -54,12 +54,6 @@ class Host(SchedulingItem):
     ok_up = 'UP'
     my_type = 'host'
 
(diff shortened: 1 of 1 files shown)


olivierHa:
Nice catch !

PR #1706, opened 2015-08-27
Service excludes/includes/overrides extension

geektophe:
This patch extends the service excludes (https://github.com/naparuba/shinken/pull/1167), includes (https://github.com/naparuba/shinken/pull/1480) and overrides (https://github.com/naparuba/shinken/pull/981) features by allowing to select target services in `service_excludes`, `service_excludes` and `service_overrides` host attribute using `*` (meaning all services attached to the host), or using a regular expression (using the `r:` prefix) on the host's services `service_description` attribute.

Examples:

    define host {
        ...
        host_name test
        service_excludes *
        ....
    }

This example would exclude all services from host `test`

    define host {
        ...
        host_name test
        service_excludes r:^proc
        ....
    }

This example would exclude all services from host `test` which `service_description` starts by `proc`

    define host {
        ...
        host_name test
        service_overrides *,notification_period workhours
        ....
    }

This example would enforce `notification_period` to `workhours` on all services from host `test`

    define host {
        ...
        host_name test
        service_overrides r:^proc,notification_period workhours
        ....
    }

This example would enforce `notification_period` to `workhours` on all services from host `test` which `service_description` starts by `proc`

**Caution**: regular expressions are case sensitive, and a regular expression which is matched by none of the services on the host would still raise an error.

### DIFF
--- a/doc/source/07_advanced/objectinheritance.rst
+++ b/doc/source/07_advanced/objectinheritance.rst
@@ -518,6 +518,18 @@ Its syntax is:
 
 It could be summarized as "*For the service bound to me, named ``xxx``, I want the directive ``yyy`` set to ``zzz`` rather tran the inherited value*"
 
+The service description selector (represented by ``xxx`` in the previous example) may be:
+
+A service name (default)
+  The ``service_description`` of one of the services attached to the host.
+
+``*`` (wildcard)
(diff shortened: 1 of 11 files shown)


geektophe:
The test errors does not seem to be related to my patch... Is there a way to run it again ?

PR #1708, opened 2015-08-27
Fix: duplicate service from template with same name

geektophe:
This patch fixes a regression appearing when services are generated with a `duplicate_foreach` statement from a common host template, and where description differs only by the expanded value of $KEY$ macro.

In such a situation, only the lastly defined host would have its services generated.

Example:

    define host{
      ...
      host_name                      test_host
      use                            generic-host
      register                       0
    }

    define host{
      use                            test_host
      host_name                      test_host_1
      _sdescr1                       svc1, svc2
    }

    define host{
      use                            test_host
      host_name                      test_host_2
      _sdescr2                       svc3, svc4
    }

    define service{
      ...
      host_name                      test_host
      service_description            $KEY$
      servicegroups                  servicegroup_01,ok
      use                            generic-service
      check_command                  command1
      duplicate_foreach              _sdescr1
      register                       0
    }

    define service{
      ...
      host_name                      test_host
      service_description            $KEY$
      servicegroups                  servicegroup_01,ok
      use                            generic-service
      check_command                  command2
      duplicate_foreach              _sdescr2
      register                       0
    }

In the previous example, only half the services would have been generated.


olivierHa:
When a "duplicate_foreach" service will be indexed in this case ?

How about indexing all services/hosts at the very end of the configuration ? (where everything has been evaluated/linked/excluded ... )

what do you think about it ?

geektophe:
In fact, `duplicate_foreach` based services do not need to be indexed (in fact, they even SHOULD NOT be indexed) as they are not real services, but stubs to instantiate new final objects.

They are trashed once expansion to real services is done. It's the resulting services which are indexed.

If they are indexed, the deduplication algorithm in `manage_conflict` method trashes the duplicates before they are expanded, leading to the bug described in the linked ticket.

geektophe:
Another point, indexing should be done as soon as possible to avoid having to walk through lists of hosts, services, contacts and so on... on every object search, which is very slow, especially on big configs where cardinality of host/services combinations may be important.

geektophe:
Sorry for this big merge, there was a conflict because of a previous PR.

PR #1710, opened 2015-08-28
Fix: fixed svc template level defined description

geektophe:
This patch fixes a series of regressions that prevented to define services which `services_description` attribute is inherited from an upper level template.

    define hostgroup {
	hostgroup_name  cisco-routers
	alias           Cisco Routers
    }

    define host{
	use         router-host
	host_name   r1
	address     1.1.1.1
	hostgroups  +cisco-routers
    }

    define host{
	use         router-host
	host_name   r2
	address     1.1.1.2
	hostgroups  +cisco-routers
    }

    define service{
	name                    ssh-critical-service
	use                     generic-service
	service_description     SSH
	register                0
    }

    define service {
	    use             ssh-critical-service
	    hostgroup_name  cisco-routers
    }

This example initially raised an error because services generated from hostgroup `cisco-routers` did not define the `service_description` attribute (this syntax is obviously correct, it should have been inherited from `ssh-critical-service`) and item pre-add checks were too restrictive.

The first fix tried to introduce the partial service notion for services not defining their `service_description`, but saving the service in an alternate dictionary broke the hostgroup explosion in the same situation: the object holding the hostgroup was saved in the `partial_services` dictionary then reintroduced in the `items` dictionary whereas it should have been trashed.

The patch removed the *partial services* related code and reworked the restrictions in `Service.add_item()` method controlling objects add and indexing.


geektophe:
Could someone run Scrutinizer again ?

olivierHa:
I don't know how to.
But I think it is safe to merge even with Scrutinizer in failed test no ?

geektophe:
I think so, there's not much changes :) I let @naparuba decide.

geektophe:
Fixed conflicts with previous PRs.

naparuba:
Let's go :)

PR #1729, opened 2015-10-16
Update Shinken documentation for Web UI new version

mohierf:
- updated WebUI screen captures
- updated pages where WebUI is mentionned
- removed webui from reference


olivierHa:
Thanks !

PR #1732, opened 2015-10-19
Push some more internal metrics into statsd

geektophe:
This patch adds (and extends) metrics sent to kernel.shinken.io and to statsd, such as number of elements in the  queues, number of managed objects, and so on... The changes are described below.

**Metrics collection**
Those metrics are collected and sent to statsd by a dedicated thread, with frequency controlled by the new `statsd_interval` parameter in the global configuration file (it defaults to 5s).

**Metrics name normalization**
The metric names have been normalized in order to have something more usable.  Each metric is made of a  base name, which may be decorated using a prefix, or a pattern string.

*Metric base name*
Metrics names follow the pattern below (parts between brackets are facultative):

    {namespace}.{service}.{name}[.{type}][.{direction}]

- The `namespace` is what the metric is related to. For instance `loop` for scheduler recurrent work methods, `tick` for tick timings, `core` for core metrics.
- The `service` is the metric emitter service. May be `arbiter`, `broker`, `poller`, `receiver`, `reactionner`, and `scheduler`. This allows to avoid collisions when metrics are emitted from services running on the same host.
- The `name` is what the metric refers to.
- The `type` is used to have variation around a common metric.
- The `direction` indicates if the metric corresponds to an input or an output stream (for counters, for instance).

Example:

    core.scheduler.hosts
    core.poller.actions.in
    hook.tick

*Metrics name decoration*
The metric names may be decorated either by a static prefix (current behaviour), or using an expansion string.

The prefix is prepended, with the host name, to the base metric.

For instance:

    statd_prefix = shinken

Leads to metrics looking like

    shinken.mon-XX.core.scheduler.hosts

The pattern allows to chose which elements are used to build the metric name.

Example:

    statsd_pattern = shinken.{host}.{metric}

Leads to the same result as below. See the documentation for a list of available macros.

Note that if both prefix and pattern as configured, pattern has precedence.

**Metrics filtering**

Each metric is set a type (`system`, `perf`, `queue`, ...).  The new parameter `statsd_types` allows to filter the metrics to be send to statsd. By default, every metrics are sent, but if this parameter is set, only metrics holding types referenced are pushed. See the documentation for a lis of metrics, and their type.

**New metrics**

See documentation for a list of available metrics, their type and a their description.

### DIFF
--- a/doc/source/03_configuration/configmain-advanced.rst
+++ b/doc/source/03_configuration/configmain-advanced.rst
@@ -1381,6 +1381,7 @@ Example:
 
 This is the api_key/scret to exchange with shinken.io and especially the kernel.shinken.io service that will print your shinken metrics. To enable it you must fill the api_key and secret parameters. You must register to http://shinken.io and look at your profile http://shinken.io/~ for your api_key and your secret.
 
+.. _configuration/configmain-advanced#statsd:
 
 Statsd host
 -----------------------------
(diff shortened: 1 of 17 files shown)


geektophe:
Please wait before merging. I have some more metrics to add.

auvipy:
good move

geektophe:
I think I added what I wanted. It should be safe to merge if you're OK with it.

mohierf:
Few doc on the most relevant metrics will be interesting ... ?

geektophe:
You're right, I'll try to add this quickly :)

geektophe:
Please wait before merging. I have another thing to check...

olivierHa:
Any updates on this ?

geektophe:
Yes, I'm working on it today, I hope it will be OK this evening :)

Sorry for having been that long, but I started this to track down memory leak issues under heavy load. And I I think I got something :) (PR will follow).

I'll let you know as soon as I've added the last metrics I was missing.

olivierHa:
Cool :)

2016-01-15 11:52 GMT+01:00 Christophe Simon <notifications@github.com>:

geektophe:
I think I'm done. Finally I slightly refactored the way I gather metrics, and normalized metrics names. I'm updating the PR description in accordingly.

naparuba:
Impressive, congrats :)

yes 4

olivierHa:
Thanks, great work !

PR #1740, opened 2015-10-27
Fix initial_state mapping for hosts

rledisez:
The initial_state "u" is declared twice, for UP and UNREACHABLE. Documentation
says that UP state is "o".


geektophe:
Oops, my bad. It's totally legit :)

olivierHa:
Thanks

PR #1746, opened 2015-11-03
HTML mail notifications fix and improvements

mohierf:
- fix #1745 exception
- add an option to specify the mail sender (-s me@myserver.com) instead of default one
- use WebUI2 company logo if defined, or default logo if not
- use WebUI2 port for webui url
- add an option to specify the WebUI root part of URL
- updated doc

PR #1749, opened 2015-11-08
notify_by_email improvements

saez0pub:
Hello,

I'm using a smtps server with login auth, so notify_by_email does,'t work.
I alreday have ssmtp which forwards the system mails. It provides a sendmail command line compatible whith the real sendmail.
Updating the mail script was necessary.
More, the sender wan be configurable.

Jean Prat.

### DIFF
--- a/libexec/notify_by_email.py
+++ b/libexec/notify_by_email.py
@@ -382,7 +382,7 @@ if __name__ == "__main__":
     else:
         sendmail = '/usr/sbin/sendmail'
         logging.debug('Send the mail')
-        p = os.popen('/usr/sbin/sendmail -t' , 'w')
+        p = os.popen('%s -t' % sendmail, 'w')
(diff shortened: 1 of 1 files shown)


olivierHa:
I've merged another PR with a sender option. Could you update this PR according to new master upstream ?

hvad:
Why include that on shinken core ? I think you must use mod -> https://shinken.readthedocs.org/en/latest/14_contributing/modules.html.

olivierHa:
hvad, I think this is the right "space"
I think we should provide at least a "decent" mail sender in shinken core. 
I agree that we can't support every "alternative" (irc,jabber ....) notification system, but email notification should be included and supported in shinken "core".

mohierf:
@saez0pub : I included your sendmail modifications in the #1746

olivierHa:
This has been merged within another PR. Thanks !

PR #1753, opened 2015-11-12
Add optional sender argument

dekonnection:
This small modification allows to optionally specify a sender address to override the default generated one (usually root@hostname) which is not always valid or pretty.

Usage : 
notify_by_email.py [...] -s "user@emailprovider.com" [...]
or
notify_by_email.py [...] --sender "user@emailprovider.com" [...]


mohierf:
For the mergers, this modification in included in #1746 ...

olivierHa:
I am closing, I've merge #1772 with is identical.

PR #1757, opened 2015-11-20
Docs - Remove the shinken-admin reference - not available in 2.x

seedickcode:
Doc cleanup - I went down the path of trying to get shinken-admin working, and debugging pyro only to find out later that shinken-admin has been deprecated / not updated to work with 2.x


olivierHa:
Thanks

JulienPalard:
Ohh that's why I can't find shinken-admin :(

PR #1766, opened 2015-11-25
Add info about exit error code when check finished with error/signalled

matkor:
Extends Action.stderrdata with check process return code information (so it's becomes visible in WebUI)
Fixes for scenario:
When check fails only by setting error code, no providing info on stderr, information about error code is lost. Similar situation is when check segfaults.
Tested on real segfaulting check scenario.


naparuba:
Thanks for the patch :)

PR #1768, opened 2015-11-26
pickle-retention-arbiter does not exist pickle-retention-file does

arthurlogilab:


naparuba:
True, but I prefer to remove the whole line, as using retention for arbiter is just too dangerous I think until there are good option to limit the different queues.

PR #1771, opened 2015-12-05
travis-ci.org - Allow Shinken to run on new infra

daniellawrence:
Also cache python packages, to speed up the builds.


olivierHa:
Thanks

PR #1772, opened 2015-12-06
included option to set sender email

fakhruddinhamid:
I have a requirement to change the sender address for the notification mail and could see on internet that many people have this need.
I included the option to set the sender field via arguments or default would be system user.


olivierHa:
Thanks

PR #1654, opened 2015-06-05
Web UI : some modifications in Shinken core are necessary to help WebUI features

mohierf:
- hostgroup_members property found in configuration is not stored in hostgroup object
- same for servicegroup_members propery ...

With those properties, the WebUI is able to build links between groups ...

PR #1776, opened 2015-12-11
Time templates and groups members

### LINKED ISSUE #1748: time_templates.cfg uses old nagios definition
The `time_templates.cfg` file uses old nagios property `normal_check_interval` ... it should be better to use `check_interval` to avoid confusing about this?

Why not include also some hosts templates in this file ... it would be interesting.

@claneys : you did include this file in many checks packs ... now it should be removed ?

mohierf:
**1/** Fixes #1748 and add hosts time templates.

**2/** Hosts, services and contacts groups members correctly managed for the Web UI (replaces old #1654)

:/P Sorry, this PR includes two different thing ...


olivierHa:
Thanks

PR #1778, opened 2015-12-17
Enh/Fix: remove BrokHandler from logging when within a worker process.

gst:
This prevent a memory leak in any module workers which use logging.

### DIFF
--- a/shinken/worker.py
+++ b/shinken/worker.py
@@ -45,10 +45,10 @@ import traceback
 import cStringIO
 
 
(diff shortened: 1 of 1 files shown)


gst:
travis build failure is a temporary failure because of port not available..

gst:
> travis build failure is a temporary failure because of port not available..

any one to restart it ?

olivierHa:
Thanks !

PR #1781, opened 2015-12-21
Use first SOFT result timestamp as timestamp of first notification

paulotfilho:
In the company I work for, we opted using the datetime of the first SOFT check result obtained for the PROBLEM notification datetime, as this might be the real datetime for the start of the problem detected. 

As a matter of fact, Nagios and Centreon already take the first SOFT result obtained as the PROBLEM notification datetime.

In order to do that, I've added an attribute called "time_first_soft_result" so I can store the timestamp of the first SOFT result of a host. On shinken/objects/schedulingitem.py, Shinken updates the attr at the first NON-OK result, which initiates the SOFT-sequence. The attribute is then used to create the first notification with the timestamp of first SOFT result.


paulotfilho:
Travis build failed because a port was not available..

PR #1783, opened 2015-12-23
create an uuid for object id, instead of increasing in int

olivierHa:
First attempt to solve the issue about missing host in livestatus.


olivierHa:
Lets merge !

Issue #1777, opened 2015-12-14
Unreachable notifications


clyra:
Hi,

I'm trying to enable unreachable notifications, but it's not working. For testing purposes i did a very simple setup with two hosts only where "host1" is parent of "host2". On both hosts definitions I have "notification_options d,u,r,f,s", and on contact definition "host_notification_options d,u,r,f,s", so it should work. No template was used, so, it shouldnt be a problem of a variable value being overrided. When both hosts went down, i got the "DOWN" notification for host1 and nothing for host2. Thruk interface shows the hosts as expected: host1 down and host2 unreachable. 

I did a clean install of shinken 2.4.2 and 2.2. Both versions displays the same behaviour. I also noticed that when host2 is on unreachable state its "duration" is being reseted at each check interval. So, lets say that host1 is down for 10 minutes, one should expect that the host2 unreachable duration is close to that, but it is not. It shows something between 0 and 1 minute (1 minute is my check interval).

Minimal configuration attached as config.zip.

ianling:
That's how it's intended to work. If the parent goes down, any of its children will also go down. So when host1 goes down, Shinken is able to determine based on your parent/child configuration that host2 will go down as well. Since it is implied that if a parent goes down, so will its child, Shinken suppresses notifications for the child.

clyra:
Suppress the child notification is the "default" behaviour but you should be able to modify it. Here[1] and here[2] says that you may choose to receive unreachable notifications using the "notification_options" and "host_notification_options", which shinken seems to ignore.

[1] http://shinken.readthedocs.org/en/latest/08_configobjects/host.html
[2] http://shinken.readthedocs.org/en/latest/08_configobjects/contact.html

paulotfilho:
This issue can be fixed by adding an "if" condition to the end of "is_no_action_dependent" function in shinken/objects/schedulingitem.py. This function uses dep status info to define if actions can be raised or not for a host (no_action boolean variable), and the default behaviour is to disable actions if all parents are dead. I added an "if" condition testing for 'u' (unreachable) in notification options. If 'u' is in notification options, the function returns False when all parents are down, and True if 'u' is not in notification options.

geektophe:
I'll have a look at this.

olivierHa:
Fix has been merged. Feel free to reopen if you encounter issues.

PR #1811, opened 2016-02-05
Fix for issue #1777

paulotfilho:
This is a fix for issue #1777, which describes the scenario of Shinken not sending notifications for hosts on unreachable state, even if you enable this in notification_options.

The "is_no_action_dependent" function uses dep status info to define if actions can be raised or not for a host (no_action boolean variable), and the default behaviour is to disable actions if all parents are dead. I added an "if" condition testing for 'u' (unreachable) in notification options. If 'u' is in notification options, the function returns False when all parents are down, and True if 'u' is not in notification options.

This was tested on Shinken 2.4.2 with Debian 8.

### DIFF
--- a/shinken/objects/schedulingitem.py
+++ b/shinken/objects/schedulingitem.py
@@ -437,8 +437,11 @@ class SchedulingItem(Item):
         # if a parent is not down, no dep can explain the pb
         if False in parent_is_down:
             return False
-        else:  # every parents are dead, so... It's not my fault :)
-            return True
+        else:  # every parents are dead, so... It's not my fault, unless I want to know about it anyway :)
+            if 'u' in self.notification_options:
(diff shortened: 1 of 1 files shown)


olivierHa:
Hello ,

maybe we should check if the notification_options attribute exists before calling it ?
notification_options isn't an attribute to the schedulingtitem class, but to "upper" classes.

What do you think ?

Regards

Olivier

paulotfilho:
Hi @olivierHa,

Sure. That would be more of a failproof approach to it. Added a commit with your request. I've used hasattr to do that, and in case the notification_options attribute doesn't exist, we return True for default behaviour.

Thanks for the tip!

olivierHa:
Thanks !

PR #1817, opened 2016-03-04
Fix attachment encoding.

k0ste:
From:
selection_001
```
Content-Type: text/html; charset="us-ascii"
MIME-Version: 1.0
Content-Transfer-Encoding: 8bit
```
To:
selection_002
```
MIME-Version: 1.0
Content-Type: text/html; charset="utf-8"
Content-Transfer-Encoding: base64
```


naparuba:
Thanks for the patch :)

PR #1821, opened 2016-03-09
Typo: Update some doc files.

mremy:
Correct arbier to arbiter, which was copied/pasted in some files.
One "No newline at end of file".


mremy:
For the default value of check_external_commands:
https://github.com/naparuba/shinken/blob/master/shinken/objects/config.py#L181

naparuba:
Thanks for the patch :)

PR #1826, opened 2016-03-17
Reverse dependency admin <-> admins.

bittrance:
When admins declared admin as a member, any standardised packaging (e.g. Docker) would have to include a user admin as declared by contacts/admin.cfg. With users declaring their groups, admin.cfg can simply be excluded in packaging.


naparuba:
Totally  agree ^^

Thanks for the patch.

PR #1828, opened 2016-03-21
Enh: Memory leak hunting

geektophe:
This patch introduces several features to protect services from  memory leaks in big environments, including queues throttling options, memory watchdog and graceful restart.

**Queues throttling**

In big environments where pollers or reactionners manage many concurrent actions, a synchronization effect may appear where a single poller or reactionner gets all the scheduled actions at once. The result is a single machine drawing almost all the workload, the other doing almost nothing.

This may lead to a chain reaction described below (let's illustrate it with a poller, it's the same for reactionners).
- A poller gets more actions it can manage.
- Under high load, the concurrency creates a contention on the internal queues.
- When contention lowers (because actions get executed), big results
  lists are returned to the scheduler.
- The scheduler has to deserialize many objects at once to integrate them. As the python interpreter does not correctly free the used memory, this may generate a memory leak: object are deserialized into memory, processed, then deleted, but the used memory is not really freed by the interpreter.
- Each integrated result generates a new brok, which creates huge broks lists.
- The broker fetching all the broks at once, it has to deserialize huge broks lists, involving the same effect as on the scheduler.

The higher the load, the worst the effect.

In addition, if results are too long to come back, the actions get rescheduled which aggravate the phenomenon. This may lead to machines crash due to memory exhaustion.

To prevent this situation, throttling options have been added to control the number of checks a poller or reactionner may get, the number of results it may return, and the number of broks a broker may get. They are described below.

_Poller and reactionner options_
- `max_q_size`: defines the maximum number of actions (slots) a poller (or reactionner) may hold.
- `q_factor`: same option as the previous one, but the number of available slots is calculated from the formula `workers * processes_by_worker * q_factor`, so it may automatically adapt to the number of CPU cores.
- `results_batch`: defines the maximum number of results a poller or reactionner may return at once.

_Broker options_
- `broks_batch`: defines the maximum number of broks a broker may get at once.

**Graceful restart**

Another memory leak comes when new configuration is sent from the arbiter. As the configuration is made of many objects linked by circular references, they are not properly garbage collected. To release this memory, the only solution is to restart the service.

This patch introduces the graceful restart. If enabled, when a new configuration is received by a daemon, it forks a new daemon and send it the received configuration on stdin. Next it exits to let the newly spawned instance replace it. This avoids having to wait for a new configuration after a classical restart.

Graceful restart can be enabled by setting the `graceful_enabled` to `1` in the daemon configuration.

**Memory watchdog**

To protect the machines from crashing because of memory exhaustion if the previous options are not or badly set, a memory watchdog may be activated with the `harakiri_threshold` in the services configuration. It automatically restarts the services if their memory goes above the threshold.

**Actions priority**

The patch also introduces the checks, event handlers and notifications priority. The commands may be added the `priority` attribute. When daemons ask the scheduler for actions, it will send the actions with the lowest `priority` first. This allows to guaranties that important actions will get executed first.

**Note**: by default, the options above are not set, and the services operate as usual.

**Note2**: this patch allows to manage slightly bigger configuration on a single scheduler. It is no more necessary to spawn many scheduler to spread the objects.

**Note3**: some of the internal data structure had to be changed to guarantee execution order.


geektophe:
There seem to be having a problem with the python 2.6 test suite. Could someone run it again ?

paulotfilho:
This is awesome

olivierHa:
@naparuba are we good to merge this PR ?

geektophe:
This patch runs on my production platform for almost 1 month, and it runs fine. The cluster is much more stable.

doc/source/07_advanced/internals-monitoring.rst
```diff
@@ -130,6 +135,9 @@ core.scheduler.checks.queue
 core.scheduler.checks.scheduled
   The checks count having scheduled state in the scheduler. Those checks have to be taken by a poller.
 
+core.scheduler.checks.reenabled
+  If checks are not done quick enoush, they are rescheduled to be handled by another reactionner. This is the number of reenabled actions.
```
petrus-v: typo:
enoush -> enough?

geektophe:
Could someone re-run the 2.6 test suite ?

olivierHa:
I don't know how to re-run it manually ?
Do re-trigger a build, I usually push an empty line or stuff like that ...

2016-04-26 16:22 GMT+02:00 Christophe Simon <notifications@github.com>:

geektophe:
@naparuba managed to re run the build last time. I don't particularly like empty commits just to re-run a test.

petrus-v:
@olivierHa,

As far I know with my Travis experience, If you are allowed to rerun a job you should find a refresh style logo on top right of corner in the summary box: https://travis-ci.org/naparuba/shinken/jobs/125800020

regards

olivierHa:
Nice ! Never noticed it :)

2016-04-26 18:10 GMT+02:00 Pierre Verkest <notifications@github.com>:

geektophe:
Thanks :)

naparuba:
Hi,

sorry for the delay.

I like the idea to enable limiting queues, but a bit less for the subprocess thing to solve memory problem (that is real).

In the Enterprise we did have such a problem in a daemon in 2/3weeks and we did find a way to solve it. It's link to the large memory allocation in threads (like here in new configuration unparsing).

We did solve this by calling:
import ctypes
libc6 = ctypes.CDLL('libc.so.6')
libc6.malloc_trim(0)

(of course it's a linux only solution, ut I think only linux have this issue). In fact Python is free() memory, but the glib is not really releasing it until it really need it, so you still see in in the RES of hte process.

Can you give a try with this intead of the subprocess? Just to see if the memory still increase (and then go for subprocess so) or if it solve the issue so we can just pick the queue limitation thing :)

geektophe:
Do you mean the graceful restart feature when you talk about subprocess ? Note that it's disabled by default, and only a manual activation in the daemons ini files will enable them.

I'll have a try with your workaround and let you know if memory is well released.

geektophe:
I had a test on my dev platform with an arbiter restart every 5 minutes.

Unfortunately, the scheduler memory continues to grow each time a new configuration is received, so does the broker memory. It took 50% more in a few hours. It's not that enormous, but it still increases until the service is restarted.

With the graceful restart, it remains perfectly stable.

I think this is not of a great interest for small setups, but it can make a difference on bigger ones. And once more, it's on opt-in option. You should only activate it if you know what you do.

paulotfilho:
@geektophe, what do you consider as a small setup and a big setup?

geektophe:
@paulotfilho It's not an absolute value, and it also depends on the way your cluster is configured, but I'd say environments with more than 10k checks start being big setups.

geektophe:
Could someone replay the 2.6 travis suite ?

olivierHa:
In progress

2016-05-02 16:32 GMT+02:00 Christophe Simon <notifications@github.com>:

geektophe:
Thanks @olivierHa :)

geektophe:
@naparuba Any news about this PR ? I'm working on another patch, and started to work from this branch. I'd like to know if I have to rollback to master. Thanks.

geektophe:
@naparuba, did you have time to have a look at this PR ? Thanks :)

efficks:
@geektophe @olivierHa @naparuba Do you have a plan to merge this PR?

geektophe:
I'm running this patch on my production platform for months with success.

I proposed to test it in #1849  and #1869 (yours :) ) and I'm waiting for feedback. The more people to test it, the safer the merge. I'd like to be sure it doesn't break existing installations.

shinken/satellite.py
```diff
@@ -765,6 +781,41 @@ def do_get_new_actions(self):
         statsmgr.incr('core.%s.actions.in' % _type, count, 'queue')
 
 
+    # Returns the maximim number of actions a sattelite may accept, and ask
```
dgilm: s/sattelite/satellite

geektophe:
It's been more than a year I use this patch in my production platform with great success. I just added some minor enhancements to graceful restart feature for more reliability, and merged master.

Once more, graceful restart is to be considered as an experimental feature and is disabled by default.

@naparuba @olivierHa could we consider merging it if you're satisfied with it ?

PS: I can't do a lot for coveralls error in tests, it came when I merged master...

geektophe:
\o/ @naparuba, I gave another try to your solution:

    import ctypes
    libc6 = ctypes.CDLL('libc.so.6')
    libc6.malloc_trim(0)

I didn't implement it correctly on the first try, now it works almost perfectly. Memory continues to grow when a new configuration is received (at least in the scheduler), but it's nothing compared to what it was without it!

In my opinion, graceful restart isn't still mandatory, but I think it can be useful, and I'd like to keep this feature. But I would understand if you preferred that I remove it.

Is it OK for you to merge ?

geektophe:
@naparuba @olivierHa is it OK for you ?

olivierHa:
Good for me ! Let's merge it !

geektophe:
After discussing with @olivierHa, I finally decided make the `free_memory()` routine call an opt-in option (PR https://github.com/naparuba/shinken/pull/1914). You now have to explicitly enable it in daemons configuration file using the `aggressive_memory_management` option.

tandrez:
Hello! 
We're having some difficulties when trying to apply this patch on our shinken 2.4.3 platform. Is the update command enough: `python setup.py install --update` ?

tandrez:
To be more precise, the livestatus module is not working anymore after applying this patch.
We have these errors our broker/livestatus logs:
```
[1561112740] ERROR: [broker-spare] [Livestatus Query] Error: 'Contacts' object has no attribute '__itersorted__'
  File "/var/lib/shinken/modules/livestatus/livestatus_query.py", line 346, in launch_query
    result = self.get_live_data(cs)
  File "/var/lib/shinken/modules/livestatus/livestatus_query.py", line 582, in get_live_data
    result = handler(self, cs)
  File "/var/lib/shinken/modules/livestatus/livestatus_query.py", line 411, in get_filtered_livedata
    items = getattr(self.datamgr.rg, self.table).__itersorted__(self.metainfo.query_hints)
AttributeError: 'Contacts' object has no attribute '__itersorted__'
```

PR #1831, opened 2016-03-27
Fixed omission of warn, grit, min or max from graphs if set to 0

dolfs:
String version of `Metric` does not include min or max values, even if they were present in the string passed to the constructor.

String version of `Metric` does not include values for warn, crit, min, or max if they were specified with value `0`. This happens because their inclusion in the string representation of `Metric` was conditional on their boolean-ish value (which is `false` for `0`). Changed to check for `None` instead.

String version of Metric did not collapse None values into empty semi-colon delimited sections, causing input of `;;1` to result in `;1` on conversion to string. Combined with above to remove collapsed `None` values from the end as well.

### DIFF
--- a/shinken/misc/perfdata.py
+++ b/shinken/misc/perfdata.py
@@ -71,12 +71,11 @@ class Metric:
                 self.max = 100
 
     def __str__(self):
-        s = "%s=%s%s" % (self.name, self.value, self.uom)
-        if self.warning:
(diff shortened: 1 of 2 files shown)


dolfs:
Travis CI build failed due to unavailability of a/the test port. This seems a problem in the build system that I cannot correct as part of this PR.

PR #1837, opened 2016-04-12
Fix #1825 : BP rules: take into account acknowledgements

### LINKED ISSUE #1825: BP Rules - take into account acknowledgements or downtimes
Case: Create a BP rule with two services.
1 services gets into warning state.
A technician acknowledges the service.

Problem : The BP_Rule still shows the state warning

BP_Rules should be able to take into account this information when computing state information.

Other systems will compute an differente state (ex. nagvis) : warning_acknowledged with a different image.

In our case, at a minimum BP_Rules should have logic to take into account other information to compute accurate state information. This should be an optional logic enhancement as I can see cases where the BP_Rule itself should be keep the existing logic and simply put the BP_Rule globally acknowledged or in maintenance. But for the majority of cases, especially when a BP_Rule contains lots of items it becomes important.

Request : Explain how to do this, if this is implemented and not documented. Implement logic in the BP_Rule also use the downtime/ack state when computing the global state.

fpeyre:
Add a parameter for a BP rule deactivate by default : business_rule_ack_as_ok
If activate, we treat host/service acknowledged as if they are Up/Ok for the evaluation of the bp rule

### DIFF
--- a/doc/source/06_medium/business-rules.rst
+++ b/doc/source/06_medium/business-rules.rst
@@ -482,5 +482,12 @@ Imagine you want to build a consolidated service which notifications contain lin
 
 The resulting output would look like ``Down web services: link1 link2 link3 ...`` where ``linkN`` are urls leading to the problem in the WebUI.
 
+Acknowledge management
+======================
+
+By default, if we acknowledge an element, it conserves its state to determine the status of our aggregated service.
+In some cases, we want that an acknowledge element is considered as if it was in Ok/Up state for a business rule.
+
+To enable this feature, simply set `business_rule_ack_as_ok`` to ``1``..
 
 .. _ticket: https://github.com/naparuba/shinken/issues/509
--- a/doc/source/08_configobjects/host.rst
+++ b/doc/source/08_configobjects/host.rst
@@ -79,6 +79,7 @@ labels                                     *labels*
 business_rule_output_template              *template*
 business_rule_smart_notifications          [0/1]
 business_rule_downtime_as_ack              [0/1]
+business_rule_ack_as_ok                    [0/1]
 business_rule_host_notification_options    [d,u,r,f,s]
 business_rule_service_notification_options [w,u,c,r,f,s]
 snapshot_enabled                           [0/1]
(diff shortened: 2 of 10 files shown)


tomasz-kuzemko:
I'm looking forward for this feature.

What about treating downtimes as OK too? As there is `business_rule_downtime_as_ack`, one might expect that in combination with `business_rule_downtime_as_ack` it will also be treated as OK.

xkilian:
@tomasz-kuzemko One might expect that in combination with business_rule_ack_as_ok it will also be treated as OK... But no, it does not work.
acknowledgements are treated as ok on services. Downtimes are not treated as acks and also are not treated as ok. As subsequent change would need to be required to do as you request. For now, the feature works as advertised and is a requirement for BP rules that provide status on groups of independant services.

fpeyre:
I just made a quick test for treating downtime as OK too.
We can use the same idea as `business_rule_ack_as_ok `. I quickly added a new parameter, `business_rule_downtime_as_ok` and it works just the same.

For `business_rule_downtime_as_ack`, according to the documentation, it said
 "By default, downtimes are not taken into account by business rules smart notifications processing. This variable allows to extend smart notifications to underlying hosts or services checks under downtime (they are treated as if they were acknowledged)."

As I understand, and after looking quickly at the code, this parameter is used only on the step to determine if a notification must be sent or not (and in this case, with this parameter, the host/service in downtime has the same behaviour as if it was acknowledged)
But it is not considered as acknowledged by the rest of code ( and in our case , the step when a BP rule is evaluated). We can rename `business_rule_downtime_as_ack` for something less ambiguous, like `business_rule_smart_notification_downtime`.

 Anyway, this features could make the object of another pull request/issue

geektophe:
In my opinion, there are 2 different aspects.

- The first one, which is the purpose of smart notifications, is to alter the conditions under which notifications are sent in business rules. Smart notification do not alter the way the state is calculated. This is the expected behavior.
- The PR you're proposing **does** alter the state processing, as the result is to **ignore** acknowledged problems (or problems under downtime), which may bring a biased information.

Those 2 different ways to approach business rules state and notifications processing should be clearly identified through the configuration options, and in the code.

PR #1842, opened 2016-04-22
Fix #1841

### LINKED ISSUE #1841: No more tags available in service objects
Relating to this issue: https://github.com/shinken-monitoring/mod-webui/issues/492.

I investigated in the framework and I noticed that the Regenerator used by the WebUI broker did not built the tags property in the service objects. In the service initial brok, the broker receives an empty tags property in the brok ...

 I suspect that the templates / tags transformation in the objects does not take place correctly.

 I added some logs in the source code to track the tags for the services and I noticed that the specific service object `linkify_templates` function in the `service.py` file is not called on Shinken start ...

 **edit**: my bad ... this function is called but not as I supposed it should be :angry: 

**edit**: found !

mohierf:
Make a correct iteration in the items of a Services object to set the tags issued from the templates used.

### DIFF
--- a/shinken/objects/service.py
+++ b/shinken/objects/service.py
@@ -1385,7 +1385,11 @@ class Services(Items):
         for i in itertools.chain(self.items.itervalues(),
                                  self.templates.itervalues()):
             self.linkify_item_templates(i)
-        for i in self:
+
+        # Then we set the tags issued from the built templates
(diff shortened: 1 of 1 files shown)


mohierf:
Fix #1841

naparuba:
Thanks :)

PR #1843, opened 2016-04-25
Fixing port numbers in documentation

epeay:
In setting up my Shinken instance, I found that some of the port numbers in the documentation were incorrect.  I have fixed the docs to show the correct port numbers.

I believe I have followed your contribution guidelines, and isolated each change in its own commit.

Please let me know if you have any questions.

Thanks!


naparuba:
Thanks for the patch :)

PR #1847, opened 2016-05-05
Improve systemd units reliability

NicolasLM:
* KillMode and TimeoutStopExec: instead of just stopping or killing the
master process, try to stop the master process, then if it
does not stop all the sub processes, kill all the sub processes one by
one. Sub processes are tracked via cgroups.
* PIDFile: do not make systemd guess which process is the master, but
instead let Shinken tell which one it the master.
* Restart and RestartSec: if Shinken exits for a strange reason
(SIGKILL, exit with wrong exit code...) wait 30 sec and restart it.


hvad:
Hi,
Sorry for the delay. I check how it work and un push PR. 
Thanks

Issue #1854, opened 2016-06-16
Feature request: --package-only option when installing


LoveIsGrief:
Hello,

#Goal 

Only install the shinken python package without any extra actions (user creation, `/etc/shinken` creation, etc.)

# Why

I'm going about developing plugins for shinken and would like to run rests using the `Perfdata` and `Metrics` classes (most likely other classes as well). To do so I try to install shinken in a virtualenv, but that doesn't work because of all the extra work going on.

It would be nice to have and extra option e.g `--package-only` to circumvent all the extra work.

Cheers

david-guenault:
you can start shinken using the start scripts from the bin folder. So it is not a realy hard thing to do.

LoveIsGrief:
That's not my goal / usecase. It's

```shell
virtualenv --system-site-packages my_shinken
source my_shinken/bin/activate
pip install shinken # --package-only
```

Which, with the current `setup.py` doesn't work.

![options1](https://cloud.githubusercontent.com/assets/2829538/16130152/2577cb96-3408-11e6-887e-b92245d7b6be.png)

PR #1856, opened 2016-06-22
Reorganize setup.py a little and add `--package-only` install option

ghost:
Hello,

This is realated to naparuba/shinken#1854 Feature request: --package-only option when installing

It is the implementation + a reorg of the setup.py code into functions. Furthermore it adds a `Vagrantfile` for easier testing of the installation via `sudo pip install /vagrant`.

A few things to mention:
- The order of the function definitions hasn't been optimized for readability
- The `main` function is now shorter in an attempt to make the logic more readable
- `pip install -e .` now works
- In order to fully install shinken after a `pip install --install-option="--package-only" .` a call to `pip install --upgrade .` should be made as it will check if shinken has already been installed
- `pip install --upgrade shinken` doesn't seem to have ever passed in the old `is_update` path, so this reorganized code doesn't either
- tests of `setup.py` are still manual but could be automated using vagrant (not done yet)

I'm very open to constructive criticism and wouldn't mind restructuring code or (if necessary) adding the tests using vagrant as mentioned above.

Cheers


naparuba:
I'll look at conflicts.

geektophe:
Hello,

This is quite interesting. Could you please rebase your branch so that it can be merged ?

Many thanks !

PR #1867, opened 2016-08-25
Fix #1866: multiple notification way when using a contact template

### LINKED ISSUE #1866: Multiple notification send when contact are in the same group and use generic-contact template
Hello,
I have an issue with multiple notification on Shinken 2.4.3
I have 2 contacts who use 2 differents notification commands. This contacts use the default  generic-contact template and are in a same contact groups. When a notification is rised, 6 notications are send : 

Here is the log from scheduler for a host problem
```
[1471628293] HOST NOTIFICATION: testUser2;172.17.0.25;DOWN;notify-host-by-email1;CRITICAL - Host Unreachable (172.17.0.25)
[1471628293] HOST NOTIFICATION: testUser2;172.17.0.25;DOWN;notify-host-by-email2;CRITICAL - Host Unreachable (172.17.0.25)
[1471628293] HOST NOTIFICATION: testUser2;172.17.0.25;DOWN;notify-host-by-email;CRITICAL - Host Unreachable (172.17.0.25)
[1471628293] HOST NOTIFICATION: testUser1;172.17.0.25;DOWN;notify-host-by-email1;CRITICAL - Host Unreachable (172.17.0.25)
[1471628293] HOST NOTIFICATION: testUser1;172.17.0.25;DOWN;notify-host-by-email2;CRITICAL - Host Unreachable (172.17.0.25)
[1471628293] HOST NOTIFICATION: testUser1;172.17.0.25;DOWN;notify-host-by-email;CRITICAL - Host Unreachable (172.17.0.25)
```
It looks like each contacts who use the same template obtain the notification command from other contacts who use the same template. The problem is gone when we don't use the template generic-contact. 

Here is the configuration I used to make this test (generic-host and  generic-contact are the default template from Shinken)

```
define host{
        use                     generic-host
        host_name               172.17.0.25
        address                 172.17.0.25
        contact_groups          testGroup
}

define contact{
    use                             generic-contact
    contact_name                    testUser1
    email                           test1.com
    host_notifications_enabled      1
    service_notifications_enabled   1
    host_notification_period        24x7
    service_notification_period     24x7
    host_notification_commands      notify-host-by-email1
    service_notification_commands   notify-service-by-email
    host_notification_options       d,u,r,f,s
    service_notification_options    c,w,r
}

define contact{
    use                             generic-contact
    contact_name                    testUser2
    email                           test2.com
    host_notifications_enabled      1
    service_notifications_enabled   1
    host_notification_period        24x7
    service_notification_period     24x7
    host_notification_commands      notify-host-by-email2
    service_notification_commands   notify-service-by-email
    host_notification_options       d,u,r,f,s
    service_notification_options    c,w,r
}

define command {
    command_name    notify-host-by-email1
    command_line    $PLUGINSDIR$/notify_by_email.py -n host -S localhost -r $CONTACTEMAIL$ -f html -c '$NOTIFICATIONTYPE$,,$HOSTNAME$,,$HOSTADDRESS$,,$LONGDATETIME$' -o '$HOSTSTATE$,,$HOSTDURATION$'
}

define command {
    command_name    notify-host-by-email2
    command_line    $PLUGINSDIR$/notify_by_email.py -n host -S localhost -r $CONTACTEMAIL$ -f html -c '$NOTIFICATIONTYPE$,,$HOSTNAME$,,$HOSTADDRESS$,,$LONGDATETIME$' -o '$HOSTSTATE$,,$HOSTDURATION$'
}

define contactgroup{
    contactgroup_name   testGroup
    alias               testGroup
    members             testUser1,testUser2
}
```
I will try to write a test for this case

fpeyre:
This PR fix a problem when 2 or more contacts ( C1, C2, ...)  with information about notification (notification period, notification commands,...) use the same contact template with a notification way. 

The problem is C1 contain notification way from the template, from C2 and from itself (Same problem for C2).

### DIFF
--- a/shinken/objects/contact.py
+++ b/shinken/objects/contact.py
@@ -297,4 +297,5 @@ class Contacts(Items):
                 if not hasattr(c, 'notificationways'):
                     c.notificationways = [nw_name]
                 else:
+                    c.notificationways = list(c.notificationways)
                     c.notificationways.append(nw_name)
--- a/test/etc/notif_way/commands.cfg
+++ b/test/etc/notif_way/commands.cfg
@@ -27,6 +27,16 @@ define command{
     #command_line    sleep 1 && /bin/true
 }
 define command{
+    command_name    notify-host-work
+    #command_line    sleep 1 && /bin/true
+    command_line    $USER1$/notifier.pl --hostname $HOSTNAME$ --notificationtype $NOTIFICATIONTYPE$ --hoststate $HOSTSTATE$ --hostoutput $HOSTOUTPUT$ --longdatetime $LONGDATETIME$ --hostattempt $HOSTATTEMPT$ --hoststatetype $HOSTSTATETYPE$
+}
+define command{
+    command_name    notify-service-work
+    command_line    $USER1$/notifier.pl --hostname $HOSTNAME$ --servicedesc $SERVICEDESC$ --notificationtype $NOTIFICATIONTYPE$ --servicestate $SERVICESTATE$ --serviceoutput $SERVICEOUTPUT$ --longdatetime $LONGDATETIME$ --serviceattempt $SERVICEATTEMPT$ --servicestatetype $SERVICESTATETYPE$
+    #command_line    sleep 1 && /bin/true
+}
+define command{
     command_name    check_service
     command_line    $USER1$/test_servicecheck.pl --type=$ARG1$ --failchance=5% --previous-state=$SERVICESTATE$ --state-duration=$SERVICEDURATIONSEC$ --total-critical-on-host=$TOTALHOSTSERVICESCRITICAL$ --total-warning-on-host=$TOTALHOSTSERVICESWARNING$ --hostname $HOSTNAME$ --servicedesc $SERVICEDESC$
 }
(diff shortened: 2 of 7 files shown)


shinken/objects/contact.py
```diff
@@ -297,4 +297,5 @@ def explode(self, contactgroups, notificationways):
                 if not hasattr(c, 'notificationways'):
                     c.notificationways = [nw_name]
                 else:
+                    c.notificationways = list(c.notificationways)
```
geektophe: Well caught ! The problem comes from `c.notificationways` being a reference on the template property, shared across objects. When values are added in the second object, they are in fact added into the root (shared) list instance.

But I think this problem is more general and should be fixed directly in https://github.com/naparuba/shinken/blob/master/shinken/objects/item.py#L325.

olivierHa:
@geektophe do you want to try to fix it from a global way ? Or too much side impact, and we can already still fix this issue this way.

geektophe:
I'll have some tests and let you know.

naparuba:
Let take this one for the moment :)

PR #1872, opened 2016-10-04
specifie host inheritance for contact/contactgroup in service definition documentation

Dodofox:
Hi,

I just learn that if contact and contact_group are not defined in service definition (and its template), it's the contact or contact group defined in host definition which is used by inheritence. 

I learned that on another page (object inheritence). I think it's better to write in this page too. 

Sorry for my english, i think you should correct my english. 

Bye


naparuba:
No that's great, thanks :)

Issue #74, opened 2011-11-14
shinken.sh for debian like (ignore gpg check)


david-guenault:
apt-get install -y should be apt-get install --force-is

PR #1882, opened 2016-12-07
Fix mod-livestatus issue #74

ianling:
mod-livestatus was unable to retrieve a contact's host notification period and service notification period, and consequently, was also unable to determine if a timestamp fell within those periods.

This fix uncomments the two linkify calls in contact.py, which allows livestatus to get the notification periods from the Contact object.

### DIFF
--- a/shinken/objects/contact.py
+++ b/shinken/objects/contact.py
@@ -227,8 +227,8 @@ class Contacts(Items):
     inner_class = Contact
 
     def linkify(self, timeperiods, commands, notificationways):
-        # self.linkify_with_timeperiods(timeperiods, 'service_notification_period')
-        # self.linkify_with_timeperiods(timeperiods, 'host_notification_period')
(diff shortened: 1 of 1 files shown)


ianling:
Not sure why Travis is failing to build this; I'm running this change in production without any issues.

geektophe:
Those calls have been commented by @naparuba in 9ae1db696704937ddff349460bdbce3fe8f49a22, I don't clearly know why. But it seems there was a reason. I'll try to find a corresponding ticket.

geektophe:
I don't see any ticket about this issue. @naparuba do you remember why it has been commented ?

naparuba:
That's because theses objects are now used not by the contact itself but by the inner notification way. So the problem is to let link just for livestatus, because the framework do not need it.

The good solution should be to find the inner notification in livestatus and get the value from it, directly in the livestatus module, so not all the framework will have theses properties linked with objects for just one module.

PR #1890, opened 2017-01-16
Add instructions to use --install-scripts in pip

clonedagain:
At least on Ubuntu Precise and Xenial, ``pip install shinken`` does not install the executables in the search path.
We have to add ``--install-option="--install-scripts=/usr/local/bin"`` to make it install to the expected directory.


naparuba:
Thanks :)

PR #1905, opened 2017-04-28
Enh: extend duplicate_foreach to host/hostgroups

geektophe:
This patch moves the phase where `duplicate_foreach` services are expanded in order to open the use of `duplicate_foreach` clause to hosts and hostgroups bound services, not only to service templates.

The parameter can be configured on services directly attached to hosts and hostgroups:

    define service {
        use                    generic-service
        service_description    Disk %KEY%
        command                check_disk!$KEY$
        duplicate_foreach      _disks
        host_name              test_host_1
        hostgroups             test_hostgroup_1
    }

### DIFF
--- a/shinken/objects/service.py
+++ b/shinken/objects/service.py
@@ -820,6 +820,9 @@ class Service(SchedulingItem):
                     continue
                 value = key_value['VALUE']
                 new_s = self.copy()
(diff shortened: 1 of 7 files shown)


naparuba:
Oh didn't remember that the duplicate was only for template based :)

PR #1914, opened 2017-06-11
Enh: Made memory free an opt-in option

geektophe:
PR https://github.com/naparuba/shinken/pull/1828 added a routine to explicitly free memory when a new configuration is loaded.

As I'm not totally sure that freeing memory using low level `libc6` calls will work on all distributions, I prefer to let the user decide whether or not to use it.

The call to `memory_free()` is now disabled by default, and may be activated by setting the `aggressive_memory_management` option to `1` in the daemons ini configuration file.

For instance, in `schedulerd.ini`:

    ...
    port=7768
    daemon_enabled=1
    aggressive_memory_management=1
    ...

### DIFF
--- a/doc/source/03_configuration/configmain.rst
+++ b/doc/source/03_configuration/configmain.rst
@@ -651,3 +651,19 @@ Default:
 
 
 Maximum delay to wait for parent configuration before giving up in a newly gracefully spawned daemon. This option has to be set in the daemon local \*d.ini files.
+
+Aggressive memory management
+----------------------------
+
+::
+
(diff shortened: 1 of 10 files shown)


geektophe:
@olivierHa if you have time to have a look at this one also :)

olivierHa:
I am curious, on which OS this option is really needed. Or maybe, it is Os specific, but "configuration" specific (many business_roles, specific module ?).
Anyway, great job !

geektophe:
Both my dev and production platforms run on ubuntu trusty, and I have many huge business rules, indeed.

Issue #1907, opened 2017-04-29
Error "ascii codec can't encode..." with "shinken search" command


Orabig:
From a fresh install of shinken : 
`# shinken --version
shinken 2.4.3
`
I was wondering why `shinken search all | grep XXX` didn't work.
Then, I got it when I tried `search all` alone : 

`# ./shinken.sh search all
3par (shirleda) [pack,3par,snmp,ssh] : System checks for 3par storage array
(...)
emc-vnx2 Traceback (most recent call last):
  File "/usr/bin/shinken", line 541, in <module>
    main()
  File "/usr/bin/shinken", line 537, in main
    CLI.one_loop(command_args)
  File "/usr/bin/shinken", line 317, in one_loop
    f(*cmd_args, **cmd_opts.__dict__)
  File "/var/lib/shinken/cli/shinkenio/cli.py", line 231, in do_search
    print_search_matches(matches)
  File "/var/lib/shinken/cli/shinkenio/cli.py", line 214, in print_search_matches
    cprint('(%s) [%s] : %s' %  (user_id, ','.join(keywords), description))
  File "/usr/local/lib/python2.7/dist-packages/shinken/misc/termcolor.py", line 134, in cprint
    print((colored(text, color, on_color, attrs)), **kwargs)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xb2' in position 74: ordinal not in range(128)
`

So the conclusion is that `shinken search all` is stopping at the `emx-vnx2` package (and thus doesn't show the following packages in alphabetical order).

The fact is that "u\xbs2" is the unicode character for 'SUPERSCRIPT TWO' ( ² ) which is part of the package name.

Unfortunately, I'm new at shinken and don't know how to fix it.

geektophe:
I had similar problem, but only when I piped the command to grep. I'm using an UTF-8 color enabled.

It seems it's a bug in the terminal detection. I'll have a look.

geektophe:
This should be fixed by #1917 Could you have a test with it ? (at least port the patch into your installation, it's straightforward).

PR #1917, opened 2017-06-22
Fix: fixed unicode string parisng in shinken cmdline

geektophe:
The `cprint` function in the `termcolor` library in the shinken cli does not correctly handle unicode strings, and raises an UnicodeEncodeError. See https://github.com/naparuba/shinken/issues/1907 for details.

This patch checks and encodes unicode strings to prevent this bug.


olivierHa:
Thanks !

Issue #1897, opened 2017-03-25
Arbitrer dies because of a connection error during a ping


ghislainp:
This is the end of a traceback that causes arbitrer death

  File "/usr/local/lib/python2.7/dist-packages/shinken/objects/satellitelink.py", line 226, in ping
    r = self.con.get('ping')
  File "/usr/local/lib/python2.7/dist-packages/shinken/http_client.py", line 139, in get
    raise Exception('Connection error to %s : %s' % (self.uri, r))
Exception: Connection error to http://localhost:7769/ : 0

it seems http_client.py should returns a HttpException instead of a basic Exception because it the former is catched by the ping function in satellitelink.py, not the latter.

This does not explain why this exception is raised, but at least the arbitrer would not die.

geektophe:
Good catch ! This is fixed in PR #1919.

PR #1919, opened 2017-06-28
Fix: http_client exceptions management

geektophe:
This patch fixes a bug in `http_client`: when an HTTP `code != 200` is got, an uncaught exception is risen, which results in a service crash. See #1897 for details.

The `Exception` instance is replaced by an `HTTPException`, which is correctly handled. The `put()` method error management behavior has also been harmonized with other methods.


olivierHa:
Good catch !

Issue #1891, opened 2017-02-14
ERROR: [Shinken] Worker '286' exit with an unmanaged exception 


nikotor:
Hi !
I use Shinken 2.4.3 on CentOS 6.8 with Python 2.6. I have this error on the poller many hours after the Poller and the Scheduler were launched. If I restart the scheduler, the poller seems to work.
These are my event log messages : 

Poller.log 
```
[1487026745] INFO: [Shinken] [poller-master] Allocating new fork Worker: 1647
[1487026745] INFO: [Shinken] Cleaning BrokHandler <shinken.log.BrokHandler instance at 0xf10fc8> from logger.handlers..
[1487026802] ERROR: [Shinken] Worker '286' exit with an unmanaged exception : Traceback (most recent call last):
  File "/usr/lib/python2.6/site-packages/shinken/worker.py", line 234, in work
    self.do_work(s, returns_queue, c)
  File "/usr/lib/python2.6/site-packages/shinken/worker.py", line 279, in do_work
    self.launch_new_checks()
  File "/usr/lib/python2.6/site-packages/shinken/worker.py", line 175, in launch_new_checks
    r = chk.execute()
  File "/usr/lib/python2.6/site-packages/shinken/action.py", line 124, in execute
    return self.execute__()  # OS specific part
  File "/usr/lib/python2.6/site-packages/shinken/action.py", line 306, in execute__
    preexec_fn=os.setsid)
  File "/usr/lib64/python2.6/subprocess.py", line 642, in __init__
    errread, errwrite)
  File "/usr/lib64/python2.6/subprocess.py", line 1238, in _execute_child
    raise child_exception
TypeError: execve() arg 3 contains a non-string value
[1486713859] WARNING: [Shinken] [poller-master] The worker 1647 goes down unexpectedly!
```

Scheduler.log
`[1487009753] WARNING: [Shinken] 4 actions never came back for the satellite 'poller-master'.I reenable them for polling`

I didn't change my configuration and all daemons are running on the same host. Is somebody can help me ? 
Thanks !

nikotor:
The output "ERROR - Plugin Timed Out" caused the issue.

nikotor:
[...]

PR #1934, opened 2017-11-14
Strip the null character from environment variables

actatux:
This makes subprocess.Popen() happy. (issue #1891)


naparuba:
Thanks for the patch :)

PR #1943, opened 2018-02-12
Fixes #1935 -- CherryPy >= 3 required

### LINKED ISSUE #1935: Check cherrypy version
It took me 3 days to figure out why the communication between the broker and the scheduler was not working after hours of checking the code.
The OS is Ubuntu 16.04. I figured out that the apt packaged names python-cherrypy is CherryPy version 2.x. Shinken needs CherryPy version 3+. So I need the packages names python-cherrypy3 and now everything works correctly.
I suggest to add cherrypy>=3 in the requirements.txt file to inform everybody you need at least version 3 and more. Also, It would be great to have a nice warning message when starting the HTTP daemon and falling back on the old HTTP backend.
Also, I suggest to remove completly the old HTTP server that does not use cherrypy at all.

efficks:
CherryPy at least version 3 and better is required for Shinken to works properly.
To completely fix #1935, we should remove the old WSGI Backend to avoid communication problems.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-CherryPy
+CherryPy>=3
 pycurl
 importlib
 pbr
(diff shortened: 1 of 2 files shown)


naparuba:
Thanks :)

PR #1971, opened 2018-09-06
Brok queues are always empty if broks_batch is unset or set to 0 (and causes Shinken to not fully start)

danirod:
This PR introduces a commit that fixes a little error in a few `if` conditionals related to brok queues. Enhancement #1828 added support for queue throttling. Unfortunately, the following fragment of code causes issues:

https://github.com/naparuba/shinken/blob/5347de922623611f140d0797490f59901d02ead2/shinken/satellite.py#L566-L575

If broks_batch is not defined, or if it's defined to 0, `count` will be computed using the statement from line 571, which does `min(broks_batch, len(self.broks))`, which equals to `min(0, len(self.broks))`, which will always equal to 0 because `min(0, whatever)` will always be 0 as long as whatever is not negative, which I don't think it's the case here.

As a consequence of this, the broker will never be able to fetch broks from the scheduler, because "up to 0 broks will be given to the broker", so the brokers will never get the initial broks required to set up modules such as webui or livestatus. Consequently, to the final users or modules, shinken will never boot completely.

My PR changes this to the behaviour that I believe was actually the expected behaviour: if broks_batch equals to 0, just return everything in the brok queues, but throttle if the parameter is actually set.

Maybe this has never been caught because configuration files from other developers already define broks_batch to something, although I could replicate this on a clean installation of Shinken made in a virtual machine with just the mod-livestatus, mod-logstore-sqlite and mod-webui modules installed.


geektophe:
Totally legit. Good catch !

danirod:
What is the status on this pull request?

Don't get me wrong, I'm not the appropiate person to point on outdated repositories, but I find funny that having the master branch of this repository broken for at least two years, the patch that would fix everything is taking so long to even get noticed by the maintainers.

geektophe:
@naparuba or @olivierHa, could you quickly have a look at that PR ?

Many thanks !

danirod:
I see a lot of test fixes recently in the main branch of shinken. Maybe _casually_ clicking the Merge button on this PR to add one more fix to the party? :upside_down_face:

danirod:
I've updated the branch now that @naparuba has finished fixing the pipelines, just so that everything looks green. 🍏📗💚

naparuba:
Yes, now tests are ok, merge is ready ^^

Thanks a lot for the patch, it's really time for it to be merged ^^

drewby08:
It looks like the latest release is still 2.4.3.  Could you cut a release in addition to having this merged to master?

PR #1990, opened 2019-10-31
Enforced downtime state after retention load

geektophe:
There is a race condition when the retention data is dumped in the
retention backend:

- The downtime depth is calculated by incrementing or decrementing the `scheduled_downtime_depth` attribute in the `Downtime` class
- If the `update_retention_file()` thread is run while a downtime is being processed, the value stored in the retention backend may not be up to date because it's read during the `enter()` or `exit()` execution:

    dt.exit()
        ...
        STOP update_retention_file()    -> value with improper value
        ...
        dt.ref.scheduled_downtime_depth -= 1

The consequence of this particular condition is that an object state can become inconsistent when the retention data is reloaded:

- All the downtimes of the object have exited
- The `scheduled_downtime_depth` remains `> 0` because of the value stored in the backend.

This enforces the downtime state evaluation when an object state is restored from the retention backend.


geektophe:
@naparuba The travis configuration is broken. I fixed it in this PR.

PR #1996, opened 2019-12-10
Remove incorrect comma

joshhodgson:
The checks don't process correctly with commas. The check also doesn't return a comma. This appears to have been copied from old Nagios documentation which is also incorrect

### DIFF
--- a/doc/source/07_advanced/perfdata.rst
+++ b/doc/source/07_advanced/perfdata.rst
@@ -41,14 +41,14 @@ Plugins can return optional performance data in their output by sending the norm
   
 ::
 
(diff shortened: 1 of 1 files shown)


geektophe:
Hello, thank you for your contribution.

I had a look, and everywhere the performance metrics are coma separated. Could you point us where you saw it's different ?

joshhodgson:
Hi, I've just spun up a Nagios instance and you can see the default checks do not use commas
image
image
and from experience (though no screenshot because I've left the org) Shinken treats the comma as part of the performance data, making it unreadable by any metrics graphing solutions. I'll try and spin up a Shinken to show this

PR #1998, opened 2020-03-10
Update about.rst to fix a typo

Victor-D:
Update about.rst to fix a typo


geektophe:
Thanks, good catch :+1:

PR #2012, opened 2021-05-10
Enh: Force problem/impact evaluation

geektophe:
When the retention data is saved by the scheduler, only the direct objects state is saved (last check time, state, state type, ...), but not the problem/impact related attributes (the associated data structure
is too complex).

This means that when the retention data is reloaded, the objects direct state is restored, but not the problem/impact related attributes, which come back to their default value. They only get recalculated when a new check result arrives arrives for a given host/service. Additionally, a Brok is only emitted if the objects have dependencies. This means that the broker isn't aware of the attribute change until a state change is triggered.

This patch adds the ability to force problem/impact evaluation on all the objects after the retention data has been reloaded. In such a situation, a Brok should not be emitted as the broker will request initial Broks soon after. The `send_brok` is there precisely to control the Broks emission in such a situation.

This feature is disabled by default, and can be enabled by setting the global parameter `enable_problem_impacts_states_reprocessing` parameter to `1` in addition to the `enable_problem_impacts_states_change` parameter in the Shinken main configuration file:

    enable_problem_impacts_states_change=1
    enable_problem_impacts_states_reprocessing=1

Some factorization has also been made in the retention data restoration routines to ease further maintenance.

### DIFF
--- a/doc/source/03_configuration/configmain.rst
+++ b/doc/source/03_configuration/configmain.rst
@@ -195,6 +195,26 @@ Default:
 This option is used to know if we apply or not the state change when a host or service is impacted by a root problem (like the service's host going down or a host's parent being down too). The state will be changed by UNKNONW for a service and UNREACHABLE for a host until their next schedule check. This state change do not count as a attempt, it's just for console so the users know that theses objects got problems and the previous states are not sure.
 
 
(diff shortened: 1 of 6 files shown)


geektophe:
@naparuba could this be of interest in Enterprise ? Did you manage this issue differently ?

PR #2014, opened 2021-07-16
Fix: missing email header to prevent "Out Of Office" messages

Zeuh:
Signed-off-by: Yann 'Ze' Richard <yann.richard@univ-rennes2.fr>


naparuba:
Merci ^^

PR #2034, opened 2024-04-11
Fix: Typo that does not reload downtimes after retention load

maethor:
Since Shinken 2.4.4 (and exactly since 5f3525f1632c1feda744f66f48e9aa3eb72ebc91), hosts and services downtimes are lost after arbiter/scheduler restarts. 

Shinken 2.4.4 removes the retention of `in_scheduled_downtime` and uses a new method after retention load to re-compute the variable by looking at `downtimes` list. BUT it computes `is_in_scheduled_downtime` instead of `in_scheduled_downtime`, so the rest does not work (at least in mod_webui, but I can't see how it could work anywhere).

This PR fixes the typo.


geektophe:
Good catch, thank you !

